PR #5, opened 2024-04-27
Update issue templates

matthiashimmelmann:
I reckon we should create templates for bug reports and feature requests so that people have rough guidelines for contributions. Feel free to change anything about the templates.


Legersky (approved):
I have reduce the number of predefined labels since I think people would not remove those that do not fit.
Otherwise I like it thought I do not know how it exactly looks like when it is used for creating a new issue.

PR #11, opened 2024-05-08
Added examples and other small variations to docstrings

MatteoGallet:


pyrigi/framework.py
```diff
@@ -258,14 +298,21 @@ def from_graph(cls, graph: Graph, dim: int) -> None:
 
         N = 10 * len(graph.vertices())**2 * dim
         realization = {
-                vertex: [randrange(1,N) for _ in range(dim)] 
+            vertex: [randrange(1, N) for _ in range(dim)]
```
matthiashimmelmann, line 301: I find this kind of notation really hard to read in-code, even if it is PEP8. Maybe we can agree upon a different notation that is also PEP8-conforming?

pyrigi/framework.py
```diff
@@ -543,21 +637,26 @@ def stress_matrix(
         """
         raise NotImplementedError()
 
-    def trivial_infinitesimal_flexes(
-            self, pinned_vertices: Dict[Vertex, List[int]] = {}) -> List[Matrix]:
+    def trivial_infinitesimal_flexes(self,
+                                     pinned_vertices: Dict[Vertex,
+                                                           List[int]] = {}) -> List[Matrix]:
```
matthiashimmelmann, line 642: This segmentation is super weird.

matthiashimmelmann (approved):

pyrigi/framework.py
Legersky: [black](https://github.com/psf/black) formats it like this:
```python
        realization = {
            vertex: [randrange(1, N) for _ in range(dim)] for vertex in graph.vertices()
```

pyrigi/framework.py
MatteoGallet: I agree, I will improve it in a PEP8-complaint way in a next feature branch.

Legersky: black:
```python
    def trivial_infinitesimal_flexes(
        self, pinned_vertices: Dict[Vertex, List[int]] = {}
    ) -> List[Matrix]:
```

PR #12, opened 2024-05-08
Feature constructors

Legersky:
- `complete_graph` and `complete_graph_on_vertices` renamed using CamelCase since they serve as class constructors.
 - `dim` parameter removed from `Framework.__init__` since the dimension is anyway retrieved from the points of realization. The only case when this is not possible is when the empty graph is passed, then the dimension is set to 0. To get empty framework with different dimension, Framework.Empty(dim) can be used.


MatteoGallet (commented):
Test passed. I agree with inferring the dimension from  the points in  the realization.

matthiashimmelmann (approved):

PR #13, opened 2024-05-08
Doc types

Legersky:
Documentation of Vertex, Edge and Point added.
They can be linked using `` :obj:`pyrigi.data_type.Vertex` `` in general, but setting 
```python
autodoc_type_aliases = {
    "Vertex": ":obj:`pyrigi.data_type.Vertex`",
    "Edge": "Edge",
    "Point": "Point", 
    }
```
in `conf.py` does not create the link correctly. So I propose we just keep it without linking.


Legersky:
wrong branch... sorry

PR #24, opened 2024-05-09
__str__, __repr__ of Graph and Framework

Legersky:
- `__str__` simplified using vertex/edge_list which are now sorted if possible
 - `__repr__` calls `__str__` which allows to get output without `print`:
 ```python
>>> from pyrigi import Graph
>>> G = Graph.from_vertices([3, 1, 7, 2, 12, 3, 0])
>>> G
Graph with vertices [0, 1, 2, 3, 7, 12] and edges []
```


Legersky:
wrong branch, again :/

PR #28, opened 2024-05-09
Method categories

Legersky:
Functionality for automatic generation of lists of method categories using decorator `@doc_category(<category-name>)`

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -79,7 +79,7 @@ napoleon_use_ivar = False
 napoleon_use_param = True
 napoleon_use_rtype = True
 napoleon_preprocess_types = True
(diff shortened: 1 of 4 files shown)


matthiashimmelmann (approved):
I was wondering if something like this exists. Great addition!

Legersky:
Thanks :)

matthiashimmelmann (approved):

MatteoGallet (commented):
I think this is a great feature. I would suggest to change

"modifications" -> "manipulation"

and to put the "Graph modification" immediately after the "Class methods".

MatteoGallet (approved):

PR #36, opened 2024-05-14
Alternative pinned vertices

MatteoGallet:
* Removed pinned vertices from Framework.rigidity_matrix().
* Created Framework.pinned_rigidity_matrix() where pinned_vertices can be given as an optional parameter (if not given, a "standard" pinning is chosen).
* Framework.inf_flexes() and Framework.nontrivial_inf_flexes() have pinned_vertices as optional parameter (if not given, no pinning is done).

### DIFF
--- a/pyrigi/framework.py
+++ b/pyrigi/framework.py
@@ -641,7 +641,6 @@ class Framework(object):
     def stress_matrix(
         self,
         data: Any,
-        pinned_vertices: Dict[Vertex, List[int]] = {},
         edge_order: List[Edge] = None,
     ) -> Matrix:
         r"""
--- a/pyrigi/framework.py
+++ b/pyrigi/framework.py
@@ -684,7 +684,6 @@ class Framework(object):
     @doc_category("Infinitesimal rigidity")
     def inf_flexes(
         self,
-        pinned_vertices: Dict[Vertex, List[int]] = {},
         include_trivial: bool = False,
     ) -> List[Matrix]:
         r"""
@@ -703,8 +702,6 @@ class Framework(object):
 
         Parameters
         ----------
-        pinned_vertices:
-            see :meth:`~Framework.rigidity_matrix`
         include_trivial:
             Boolean that decides, whether the trivial motions should
             be included (`True`) or not (`False`)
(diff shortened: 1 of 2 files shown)


pyrigi/framework.py
```diff
@@ -542,106 +541,130 @@ def rigidity_matrix(
...
-            edge_order = sorted(self._graph.edges())
+        if edge_order is None:
+            edge_order = self._graph.edge_list()
         else:
-            edge_order = self._graph.edges()
+            if not set([set(e) for e in self._graph.edges]) == set(
+                [set(e) for e in edge_order]
+            ):
+                raise ValueError(
+                    "edge_order must contain " "exactly the same edges as the graph!"
```
matthiashimmelmann, line 572: Is this a typo?

pyrigi/framework.py
```diff
@@ -542,106 +541,130 @@ def rigidity_matrix(
...
                 )
-
-        if not edges_ordered:
-            edge_order = sorted(self._graph.edges())
+        if edge_order is None:
+            edge_order = self._graph.edge_list()
         else:
-            edge_order = self._graph.edges()
+            if not set([set(e) for e in self._graph.edges]) == set(
+                [set(e) for e in edge_order]
```
matthiashimmelmann, line 569: I think we also want to check that no edges are there twice. Thus not only `set==set`, but also `list(set(...))==...`.

pyrigi/framework.py
```diff
@@ -542,106 +541,130 @@ def rigidity_matrix(
...
+            ):
+                raise ValueError(
+                    "edge_order must contain " "exactly the same edges as the graph!"
+                )
 
         # `delta` is responsible for distinguishing the edges (i,j) and (j,i)
-        def delta(u, v, w):
-            if w == u:
+        def delta(e, w):
```
matthiashimmelmann, line 576: I assume `e` for edge? Can you add a comment clarifying this?

pyrigi/framework.py
MatteoGallet: You're right, thanks!

pyrigi/framework.py
MatteoGallet: Yes, ok!

pyrigi/framework.py
MatteoGallet: Yes, it may have been a byproduct of the linter

test/test_framework.py
```diff
@@ -45,40 +47,13 @@ def test_inf_rigidity():
     assert F_.is_inf_rigid() and F_.is_min_inf_rigid()
 
 
-def test_inf_motions():
+def test_inf_flexes():
```
matthiashimmelmann, line 50: Please add tests for `rigidity_matrix` and `pinned_rigidity_matrix`, since there were significant changes.

matthiashimmelmann (approved):

PR #47, opened 2024-05-22
Graph/framework databases for parametrized testing

Legersky:
To avoid creating the same graphs/frameworks again and again in the tests, we can use fixtures.


Legersky:
In order to allow using graphs/frameworks that were in fixtures also outside tests, we replace the fixtures by graph/framework databases.

Legersky:
Now some of the original tests are rewritten as parametrized ones including many more graphs/frameworks from the databases, so the purpose of the branch is in my opinion achieved.

> [!WARNING]  
> Some of the frameworks in `test_inf_rigid` and `test_min_inf_rigid` are commented out because there is a bug, probably in `is_inf_rigid`: for instance $K_3$ is infinitely rigid for any injective realization in any dimension! We should fix this on a separate branch after merging this one.

MatteoGallet (approved):

PR #59, opened 2024-06-26
Methods for equivalence and congruence

rubesjan:
data types are changed, added point_to_vector function
added translation and rotation2D of a framework

### DIFF
--- a/pyrigi/data_type.py
+++ b/pyrigi/data_type.py
@@ -5,7 +5,8 @@ Module for defining data type used for type hinting.
 """
 
 from sympy import Matrix
-from typing import TypeVar, List, Tuple, Hashable
+from typing import TypeVar, Tuple, Hashable
+from collections.abc import Sequence
 
 
 Vertex = Hashable
@@ -18,13 +19,18 @@ Edge = Tuple[Vertex, Vertex]
 An Edge is a pair of :obj:`Vertices <pyrigi.data_type.Vertex>`.
 """
 
-Point = List[float]
+Coordinate = int | float | str
 """
-A Point is a list of coordinates whose length is the dimension of its affine space.
+An integer, float or a string interpretable by :func:`sympy.core.sympify.sympify`.
+"""
+
+Point = Sequence[Coordinate]
+"""
+A Point is a Sequence of Coordinates whose length is the dimension of its affine space.
 """
 
 
-Vector = TypeVar(Matrix)
+Vector = TypeVar("Matrix")
 GraphType = TypeVar("Graph")
 FrameworkType = TypeVar("Framework")
 MatroidType = TypeVar("Matroid")
--- a/pyrigi/data_type.py
+++ b/pyrigi/data_type.py
@@ -4,7 +4,7 @@ Module for defining data type used for type hinting.
 
 """
 
-from sympy import Matrix
+from sympy import Matrix, SympifyError
 from typing import TypeVar, Tuple, Hashable
 from collections.abc import Sequence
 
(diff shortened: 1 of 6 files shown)


pyrigi/data_type.py
```diff
@@ -17,11 +19,39 @@
...
+    """
+    Return point as single column sympy Matrix.
+    """
+
+    if isinstance(point, MatrixBase):
+        if point.shape[1] != 1:
+            raise ValueError("Point could not be interpreted as column vector.")
+        return point
+
+    if not isinstance(point, Sequence) or isinstance(point, str):
```
MatteoGallet, line 47: May I ask you why you have `or isinstance(point, str)` ?

pyrigi/framework.py
```diff
@@ -1082,13 +1082,194 @@ def is_redundantly_rigid(self) -> bool:
...
+            The framework for checking the congruence.
+        numerical
+            Whether the check is symbolic (default) or numerical.
+        tolerance
+            Used tolerance when checking numerically.
+        """
+
+        if not nx.utils.graphs_equal(self._graph, other_framework._graph):
+            raise ValueError(
+                "Not all vertices have a realization in the given dictionary"
```
MatteoGallet, line 1144: This error message seems a bit cryptic to me: am I right if I say that these lines check whether the underlying graphs of the two frameworks are equal? If so, I would suggest to write something like "The underlying graph of ... is different from the underlying graph of ...".

Legersky (approved):

pyrigi/framework.py
rubesjan: Yes, the message was wrong. I rewrote it to "Given graphs are not same."
The methods is_congruent_realization and is_equivalent_realization can raise the exception "Not all vertices have a realization in the given dictionary", because they are given on input the realization in the form of dictionary (and the dictionary has to have a realization for all the vertices ofc)
Thank you for the checking

MatteoGallet: Thank you! Would you mind adding "underlying", since the method applies to frameworks and not to graphs?
Indeed, realizations are dictionaries, but I believe that, for the user who might be less interested in the way how the class is implemented, it would be more transparent to refer to the mathematical object that causes the error, rather than the Python data type used to implement it.
Thanks again.

rubesjan: Added underlying to the message.
When using is_equivalent_realization or is_congruent_realization methods directly, the user have to know, that the realization needs to be represented as a dictionary, otherwise he would not be able to use these methods directly.
is_equivalent_realization method is called from is_equivalent only if the dictionary contains all the vertices, so the exception could not be raised from is_equivalent_realization

MatteoGallet: Thanks!

pyrigi/framework.py
```diff
@@ -1082,13 +1082,198 @@ def is_redundantly_rigid(self) -> bool:
...
+        """
+        Translate the framework.
+
+        Parameters
+        ----------
+        vector
+            Translation vector
+        inplace
+            If True (default), then this framework is translated.
+            Otherwise new translated framework is returned.
```
MatteoGallet, line 1231: Please change
"Otherwise new translated framework is returned." -> "Otherwise, a new translated framework is returned."

pyrigi/framework.py
```diff
@@ -1082,13 +1082,198 @@ def is_redundantly_rigid(self) -> bool:
...
+        """
+        Rotate the planar framework counter clockwise.
+
+        Parameters
+        ----------
+        angle
+            Rotation angle
+        inplace
+            If True (default), then this framework is rotated.
+            Otherwise new rotated framework is returned.
```
MatteoGallet, line 1259: Please change
"Otherwise new translated framework is returned." -> "Otherwise, a new translated framework is returned."

pyrigi/framework.py
```diff
@@ -1082,13 +1082,198 @@ def is_redundantly_rigid(self) -> bool:
...
+        inplace
+            If True (default), then this framework is translated.
+            Otherwise new translated framework is returned.
+        """
+
+        vector = point_to_vector(vector)
+
+        if inplace:
+            if vector.shape[0] != self.dim():
+                raise ValueError("Different dimension!")
```
MatteoGallet, line 1238: Could you please make this error message more explanatory?

MatteoGallet (approved):

PR #62, opened 2024-07-01
Functions to check if a list of edges is contained in a graph added to graph.py

MarketaHosmankova:


Legersky:
Could you distinguish the following cases, please?
  -  `TypeError` :  `edge` is not a `Sequence` of length 2 (`from collections.abc import Sequence`)
  - `ValueError` : `edge` is not an edge of the graph, its endvertices are not in `vertices` 
 
The following updated test should pass:
```python
def test_check_edge_list():
    G = Graph.from_vertices_and_edges([1, 2, 3], [(1, 2), (2, 3)])
    G._check_edge((1, 2))
    G._check_edge(3, 2)
    G._check_edge_list([(1, 2), (2, 3)])
    G._check_edge_list([(1, 2)], [1, 2])
    G._check_edge_list([(2, 3)], [2, 3])
    with pytest.raises(ValueError):
        G._check_edge((1, 3))
    with pytest.raises(ValueError):
        G._check_edge((1, 4))
    with pytest.raises(ValueError):
        G._check_edge_list([(1, 2), (1, 3), (2, 3)])
    with pytest.raises(ValueError):
        G._check_edge_list([(1, 2), (2, 3)], [1, 2])
    with pytest.raises(TypeError):
        G._check_edge_list([(2,)])
    with pytest.raises(TypeError):
        G._check_edge_list([2,3])
    with pytest.raises(TypeError):
        G._check_edge_list(['23'])
```

Please, add also a brief docstring describing the purpose of `vertices`.

MarketaHosmankova:
I made the two cases distinct and added the requested docstring.
In the updated tests, the test G._check_edge(3, 2) is not supposed to pass because the function expects a tuple as an argument, I rewrote it to G._check_edge((3, 2)).

Legersky:
You are right, my mistake with the `G._check_edge(3, 2)`.
But an edge given as a list should pass as well (`G._check_edge([3, 2])`), so `isinstance(edge, tuple)` should be `isinstance(edge, Sequence)`.

MarketaHosmankova:
Alright, I will edit that and add a corresponding case to the test file.

MarketaHosmankova:
Currently no function checking if a variable is an instance of a sequence is imported to the code, so I solved this issue by explicitly listing checks for a tuple and a list.

MarketaHosmankova:
For my own usage it does not matter whether the check is split into two functions or not, from reading through the file I think the functions can now do everything I need them to.

pyrigi/graph.py
```diff
@@ -155,6 +150,73 @@ def CompleteOnVertices(cls, vertices: List[Vertex]) -> GraphType:
         edges = combinations(vertices, 2)
         return Graph.from_vertices_and_edges(vertices, edges)
 
+    def _check_edge_format(self, input_pair: Edge) -> None:
+        """
+        Check if an input_pair is a pair of distinct vertices of the graph.
+        """
```
MatteoGallet, line 156: Should we add `Parameters` in the docstring?

Legersky: I think it isn't necessary since the meaning is clear from the docstring. The method is not in the documentation as it starts with _. Even if it was, the section parameters would be added automatically, see misc.is_zero_vector.

Legersky: Or do you want to describe more precisely the meaning of `input_pair` in Parameters?

MatteoGallet: Thanks, yes, maybe one could specify a bit more the meaning of `input_pair`, but on the other hand the docstring is informative enough. So it is ok for me to leave it as it is.

MatteoGallet (approved):

PR #65, opened 2024-07-04
Feature graph extensions

MarketaHosmankova:
I have added code for k_extension, all_k_extensions and extension_sequence.

The output for extension_sequence is currently a list of all graphs in the sequence, with @Legersky we talked about instead making it a list of changes between two consecutive graphs or giving the user a choice between these two, I have however not implemented this feature yet.

The corresponding tests require comparison between two graphs, which cannot currently be done directly, so I propose I could also implement a function to compare if two graphs are equal.

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -14,8 +14,18 @@ jobs:
       pages: write
       id-token: write
     steps:
+    - uses: actions/checkout@v4
+    - name: Set up Python 3.10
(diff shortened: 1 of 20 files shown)


Legersky:
NetworkX has a function to compare graphs:
https://networkx.org/documentation/stable/reference/generated/networkx.utils.misc.graphs_equal.html#networkx.utils.misc.graphs_equal

Legersky:
Please, create the pull request to `dev` instead of `main`.

PR #66, opened 2024-07-04
Feature graph extensions

MarketaHosmankova:
I have added code for k_extension, all_k_extensions and extension_sequence.

The output for extension_sequence is currently a list of all graphs in the sequence, with @Legersky we talked about instead making it a list of changes between two consecutive graphs or giving the user a choice between these two, I have however not implemented this feature yet.


MarketaHosmankova:
flake8 didn't accept GraphType as return type. I do not know how to fix this, so I set the return type to Any for now.

pyrigi/graph.py
```diff
@@ -322,57 +322,256 @@ def is_tight(self, K: int, L: int) -> bool:
...
+    def zero_extension(
+        self,
+        vertices: List[Vertex],
+        new_vertex: Vertex = None,
+        dim: int = 2,
+        inplace: bool = False,
+    ) -> Any:
         """
-        Notes
+        Return a k-extension of the graph, where k = 0.
```
Legersky, line 334: You can write simply `Return a 0-extension of the graph.`

Legersky (commented):
The only big problem I see is that the base case in an extension sequence (namely the first graph in the sequence) has to be an edge since an edge cannot be obtained from a single vertex by 0/1-extension.
Could you fix this, please?

There are also some `TypeError` to be changed to `ValueError`, see the comments.

Regarding the checks of equality via strings, I will start a new branch on implementing comparison using `==` once we merge this branch. (There are some details to be discussed so I do not want to do it here).

Could you maybe add a test of `all_k_extensions(only_non_isomorphic=True)` for (k,d) = (0,2), (1,2) on a non-complete graph, for instance Diamond?

Could you add the definition of k-extension to `doc/math/definitions.md` and link it from the docstring? I can help with that if you have troubles.

pyrigi/graph.py
```diff
@@ -322,57 +322,257 @@ def is_tight(self, K: int, L: int) -> bool:
...
-        raise NotImplementedError()
+        if self.number_of_nodes() < (dim + k):
+            raise TypeError(
+                "The number of nodes in the graph needs to be "
+                "greater or equal than the dimension + k!"
+            )
+        if self.number_of_edges() < k:
+            raise TypeError(
+                "The number of edges in the graph needs to be greater or equal than k!"
+            )
```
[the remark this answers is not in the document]
MatteoGallet: In addition, I think it would be better if the string were:
f"The number of edges in the graph needs to be greater or equal than {k}."

pyrigi/graph.py
```diff
@@ -322,57 +322,257 @@ def is_tight(self, K: int, L: int) -> bool:
...
         if not isinstance(dim, int) or dim < 1:
             raise TypeError(
                 f"The dimension needs to be a positive integer, but is {dim}!"
             )
-        raise NotImplementedError()
+        for vertex in vertices:
+            if vertex not in self.nodes:
+                raise TypeError(f"Vertex {vertex} is not contained in the graph")
+        if len(set(vertices)) != dim + k:
+            raise TypeError("List of vertices must contain dim + k distinct vertices")
```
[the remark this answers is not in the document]
MatteoGallet: In addition, I think it would be better if the string were
`f"List of vertices must contain {dim + k} distinct vertices"`

MatteoGallet (commented):
As a general comment, I would suggest to create, alongside the current implementation, the possibility of creating an iterator over all possible k-extensions of a graph. In fact, I believe it is likely that a user may be interested in iterating over such collections.

MarketaHosmankova:
Now it is also possible to get an iterator over all k-extensions of a graph.

Legersky:
I fear we have misunderstood each other with the base case of an extension sequence: a graph has an extension sequence if and only if it it can be constructed from the complete graph on 2 vertices (i.e. a single edge) by a sequence of 0 and 1-extension.
In particular, the graph formed by a single vertex does **not** have an extension sequence. An extension sequence always starts with an edge.

MarketaHosmankova:
Alright, my bad, I will fix it. However, the previous version of the code should return that no extension sequence exists for a graph with less than two vertices and one edge. Should I make it throw an error instead?

Legersky:
`False`/`None` should be returned depending on `return_solution`, not an error, since these graphs just don't have an extension sequence.

pyrigi/graph.py
```diff
@@ -322,57 +322,267 @@ def is_tight(self, K: int, L: int) -> bool:
...
+            if self.number_of_edges() == 1:
+                G = deepcopy(self)
+                G.remove_node(list(G.nodes)[0])
+                return [G, self] if return_solution else True
+            return None if return_solution else False
+        if (
+            self.number_of_nodes() < 2
+            or not self.number_of_edges() == 2 * self.number_of_nodes() - 3
+        ):
+            return None if return_solution else False
```
MarketaHosmankova, line 559: These lines will make the function return None/False if a graph with less than two vertices or one edge is passed.

Legersky:
Regarding Matteo's comment about iterating: the point of returning an iterator is to avoid keeping the whole list in memory. So the approach here would be to implement it as a generator (which is a special kind of iterator) using `yield`. This is doable for listing all extensions, but difficult for returning only nonisomorphic ones (one would have to either remember the previously returned graphs, which makes using a generator a bit pointless, or to have some trick how to generate directly only nonisomorphic ones, no idea how to achieve this...). So it is up to you whether you want to implement it using generator for `only_non_isomorphic=False`, or to remove `return_iterator` option for now.

pyrigi/graph.py
Legersky: If the condition on line 556 is true, then the one on 557 is true as well, so you can keep just the second one. But there is still some problem, since the returned extension sequences start with a single vertex (the first graph in solutions in `test_extension_sequence_solution` has to be removed).

MarketaHosmankova: I rewrote the code to return sequences starting with a complete graph on two vertices, I hope it is correct this time.

MarketaHosmankova:
I was not familiar with generators in python, but from my knowledge on k-extensions I suppose that there will be many cases where a graph has a high amount of possible k-extensions but only several of them are non-isomorphic, so I think implementing a generator for all_k_extensions would be much more beneficial for non_isomorphic_only = False than for non_isomorphic_only = True from the point of performance. I might be wrong on this though.

pyrigi/graph.py
Legersky: It looks good, thanks.

Legersky:
I tried to compare just a bit the time performance of between the implementation as list vs generator. It seems that there is no significant difference for `only_non_isomorphic=True`. My guess is that the isomorphism check is the bottleneck anyway so we cannot expect to be able to get the result in reasonable time for graphs bigger than a few vertices. :
image
image
image
image
Hence, I take back my comment about being pointless to implement it as a generator with  `only_non_isomorphic=True`, for such small graphs it does not really matter that we have to store the previously returned graphs. Generators should be definitely better regarding memory. So I would say that returning always a generator is a clean solution. Should I commit what I used for my tests?

Legersky:
I think the only missing thing is to update `doc/math/definitions.md` with k-extensions and link it from the docstrings.

MarketaHosmankova:
I updated definitions.md, although I am not sure if I added everything that should be there.

MatteoGallet (approved):

MatteoGallet (approved):

PR #68, opened 2024-07-12
Test and add examples for complete on vertices

MarketaHosmankova:
Added examples and test for the function CompleteOnVertices from graph.py.


MarketaHosmankova:
It seems like there is a problem with graph_from_vertices_and_edges because CompleteOnVertices now returns a graph with no edges.

MarketaHosmankova:
> It seems like there is a problem with Graph.from_vertices_and_edges because CompleteOnVertices now returns a graph with no edges.

Now I checked it and it looks like Graph.from_vertices_and_edges works.

Legersky:
My guess is that in `CompleteOnVertices`, `combinations(vertices, 2)` return an iterator which is consumed by the `_check_edge_format_list` so it becomes empty, but I am still puzzled.

MarketaHosmankova:
I was thinking the same, I will try to fix it by making a list of combinations. I suppose this problem didn't occur before because the iterator was sent directly to from_vertices_and_edges.

Legersky:
We can do
```
        G = Graph.from_vertices(vertices)
        G.add_edges(combinations(vertices, 2))
        return G
```
in `CompleteOnVertices`, since we know we can skip the check.

MarketaHosmankova:
I solved it by passing a list of combinations instead of an iterator.

Legersky:
Thanks a lot for discovering the bug by writing the test!

Legersky (approved):

PR #71, opened 2024-07-12
Speeding up min and max rigid subgraphs

Legersky:
The functions `min_rigid_subgraphs` and `max_rigid_subgraphs` have been rewritten to speed up the testing.

### DIFF
--- a/pyrigi/graph.py
+++ b/pyrigi/graph.py
@@ -810,7 +810,7 @@ class Graph(nx.Graph):
 
         Notes
         -----
-        We only return nontrivial proper subgraphs, meaning that there need to be at
+        We only return nontrivial subgraphs, meaning that there need to be at
         least ``dim+1`` vertices present.
 
         Examples
@@ -820,7 +820,7 @@ class Graph(nx.Graph):
         >>> G.is_rigid()
         True
         >>> G.min_rigid_subgraphs()
-        []
+        [[0, 1, 2, 3, 4, 5]]
         >>> G = graphs.ThreePrism()
         >>> G.is_rigid()
         True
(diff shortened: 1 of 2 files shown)


pyrigi/graph.py
```diff
@@ -779,41 +779,34 @@ def max_rigid_subgraphs(self, dim: int = 2) -> List[Graph]:
...
-                    break
-            if not iso_bool:
-                clean_list.append(max_subgraphs[i])
-        return clean_list
+        rigid_subgraphs = {
+            tuple(vertex_subset): True
+            for r in range(dim + 1, self.number_of_nodes() - 1)
+            for vertex_subset in combinations(self.nodes, r)
+            if self.subgraph(vertex_subset).is_rigid(dim)
+        }
```
MatteoGallet, line 789: Would it be reasonable to first determine the connected components and then run through their subgraphs, so that one avoids, in case the graph is disconnected, many situations that are surely `False`?

pyrigi/graph.py
```diff
@@ -779,41 +779,34 @@ def max_rigid_subgraphs(self, dim: int = 2) -> List[Graph]:
...
+
+        for H1, H2 in combinations(rigid_subgraphs.keys(), 2):
+            if set(H1).issubset(set(H2)):
+                rigid_subgraphs[H1] = False
+            if set(H2).issubset(set(H1)):
+                rigid_subgraphs[H2] = False
+        return [list(H) for H, is_max in rigid_subgraphs.items() if is_max]
```
MatteoGallet, line 796: Since we aim at the maximal rigid subgraphs, would it be reasonable to start from the subgraphs with maximal number of vertices and then proceed towards those with fewer vertices, pruning the branches that already have a rigid subgraphs, rather than computing all rigid subgraphs and then extracting the maximal ones?

pyrigi/graph.py
Legersky: Yes, we could do that.

pyrigi/graph.py
Legersky: I don't know how to do this easily. I think that was the idea of the original implementation with recursion, but on the single test we have this is faster.

Legersky:
(The aim of this refactoring was to get quickly something reasonably fast to speed up the tests, both methods would deserve more attention in the future.)

pyrigi/graph.py
Legersky: Or do you mean just when we are extracting the maximal ones from the list of the rigid subgraphs?

Legersky: the latter is implemented now

MatteoGallet: Thanks! Yes, I think it is worth investigating algorithms that do not need to list all rigid subgraphs, but this can be done at a later stage.

MatteoGallet (approved):

pyrigi/graph.py
Legersky: For d=2, pebble games allow to extract maximal rigid components much faster.

PR #72, opened 2024-07-13
Equality of graphs

Legersky:
Do we want to be able to check whether two graphs are equal like `G == H`? This can be done in Python by implementing method `__eq__`, see the first commit of this branch. I think it would be useful for instance in tests.
If yes, when do we consider two graphs to be equal? Same vertex and edge sets? Could it happen in future that we have for instance some edge labelings?

### DIFF
--- a/pyrigi/graph.py
+++ b/pyrigi/graph.py
@@ -104,7 +104,7 @@ class Graph(nx.Graph):
         Note
         ----
         :func:`~networkx.utils.misc.graphs_equal(self, other)`
-        behaves strange, hence it is not used.
+        behaves strangely, hence it is not used.
         """
         if (
             self.number_of_edges() != other.number_of_edges()
--- a/pyrigi/graph.py
+++ b/pyrigi/graph.py
@@ -97,6 +97,28 @@ class Graph(nx.Graph):
         """
         return self.__str__()
 
+    def __eq__(self, other: Graph):
+        """
+        Return whether the other graph has the same vertices and edges.
+
+        Note
+        ----
+        :func:`~networkx.utils.misc.graphs_equal(self, other)`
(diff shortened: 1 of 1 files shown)


MatteoGallet:
I think it is a good idea to have a check for equality. I would say that two graphs are equal when they have both vertex sets and edge sets equal. Then if at some point we implement directed/multi/labelled graphs, I think we will do that via classes that inherit from Graph, and then we will write corresponding `__eq__` methods.

Legersky:
Directed/multigraph are different object in Networkx, so we indeed do not need to worry about them now. Regarding labelled graphs, Networkx Graph from which we inherit allows graph-edge-vertex attributes: https://networkx.org/documentation/stable/tutorial.html#adding-attributes-to-graphs-nodes-and-edges
Do we want to check those for equality?

MatteoGallet:
> Directed/multigraph are different object in Networkx, so we indeed do not need to worry about them now. Regarding labelled graphs, Networkx Graph, from which we inherit, allows graph/edge/vertex attributes: https://networkx.org/documentation/stable/tutorial.html#adding-attributes-to-graphs-nodes-and-edges Do we want to check those for equality?

I see... Yes, in this case I think they should be checked. I would use an optional argument that allows one not to check them by default.

Legersky:
If we want to use arguments, we should not implement `__eq__`, but `is_equal`, since `G == H` syntax does not take any arguments.

MatteoGallet:
> If we want to use arguments, we should not implement `__eq__`, but `is_equal`, since `G == H` syntax does not take any arguments.

I see! Ok, then I would leave this as it is and in case implement another feature for `is_equal`

MatteoGallet (approved):

PR #73, opened 2024-07-31
Graph plotting

Legersky:
I have polished a bit the plotting functionality implemented by @lenkakusnirova.
Thanks to her, graphs can be plotted now with various formatting options.
The same options are available for plotting planar frameworks.

We should decide for better default colors (@ggmath ?) and whether we want to distinguish plots of graphs and frameworks somehow.

If we do not mind adding a dependency, we could use this package to have automatically generated colors that are more distinct for the option `edge_color` when a partition of edges is specified: https://distinctipy.readthedocs.io/en/latest/


MatteoGallet:
Thanks to @lenkakusnirova!

It seems to me that `distinctipy` seems a nice and light tool, I would be in favor of using it.

One aspect that we may change: the `doc_category` of the `plot` method in `Graph` is `"Other"`, while the `doc_category` of the `plot` method in `Framework` is `"Plotting"`.

pyrigi/graph.py
```diff
@@ -1399,6 +1401,172 @@ def random_framework(self, dim: int = 2, rand_range: Union(int, List[int]) = Non
...
+                        )
+                    edge_color_array.append(colors[i])
+                    edge_list_ref.append(tuple(e))
+        elif isinstance(edge_color, dict):
+            color_edges_dict = edge_color
+            for color, edges in color_edges_dict.items():
+                for e in edges:
+                    if not self.has_edge(e[0], e[1]):
+                        raise ValueError(
+                            "Input includes edge that is not part of the framework"
```
MatteoGallet, line 1436: "Input includes edge" -> "The input includes an edge"

pyrigi/graph.py
```diff
@@ -1399,6 +1401,172 @@ def random_framework(self, dim: int = 2, rand_range: Union(int, List[int]) = Non
...
+        else:
+            raise ValueError("The input color_edge has none of the supported formats.")
+        for e in edge_list:
+            if (e[0], e[1]) not in edge_list_ref and (e[1], e[0]) not in edge_list_ref:
+                edge_color_array.append("black")
+                edge_list_ref.append(e)
+        if len(edge_list_ref) > self.number_of_edges():
+            raise ValueError(
+                "There is an edge whose color was specified multiple times."
+            )
```
MatteoGallet, line 1449: Shall we provide a more refined error message, i.e, identify which edge(s) color is specified multiple times?

pyrigi/graph.py
```diff
@@ -1399,6 +1401,172 @@ def random_framework(self, dim: int = 2, rand_range: Union(int, List[int]) = Non
...
+        return edge_color_array, edge_list_ref
+
+    @doc_category("Other")
+    def layout(self, layout_type: str = "spring") -> dict[Vertex, Point]:
+        """
+        Generate a placement of the vertices.
+
+        This method a is wrapper for functions
```
MatteoGallet, line 1457: "This method a is wrapper for functions" -> "This method a is wrapper for the functions"

pyrigi/graph.py
```diff
@@ -1399,6 +1401,172 @@ def random_framework(self, dim: int = 2, rand_range: Union(int, List[int]) = Non
...
+
+        See tutorial Plotting for illustration of the options.
+
+        Parameters
+        ----------
+        placement:
+            The placement of vertices in the plane.
+        layout:
+            If ``placement`` is not specified,
+            then it is generated using :meth:`~Graph.layout`.
```
MatteoGallet, line 1508: What does "it" refer to here?

pyrigi/graph.py
```diff
@@ -1399,6 +1401,172 @@ def random_framework(self, dim: int = 2, rand_range: Union(int, List[int]) = Non
...
+        vertex_size:
+            The size of the vertices.
+        vertex_color:
+            The color of the vertices. Color can be string or rgb (or rgba)
```
MatteoGallet, line 1512: "Color can be" -> "The color can be"

pyrigi/graph.py
```diff
@@ -1399,6 +1401,172 @@ def random_framework(self, dim: int = 2, rand_range: Union(int, List[int]) = Non
...
+        vertex_shape:
+            The shape of the vertices specified as as matplotlib.scatter
+            marker, one of ``so^>v<dph8``.
+        vertex_labels:
+            If ``True`` (default), vertex labels are displayed.
+        edge_width:
+        edge_color:
+            If a single color is given as a string or rgb (or rgba) tuple
+            of floats from 0-1, then all edges get this color.
+            If a (possible incomplete) partition of the edges is given,
```
MatteoGallet, line 1523: "possible incomplete" -> "possibly incomplete"

pyrigi/graph.py
```diff
@@ -1399,6 +1401,172 @@ def random_framework(self, dim: int = 2, rand_range: Union(int, List[int]) = Non
...
+            then each part gets a different color.
+            If a dictionary from color to a list of edge is given,
```
MatteoGallet, line 1525: "from color to a list" -> "from colors to a list"

MatteoGallet (changes requested):

pyrigi/graph.py
Legersky: As you prefer, I was lazy to implement something robust enough to deal with swapped vertices when specified second time.

pyrigi/graph.py
Legersky: placement

MatteoGallet: Ok! But why is this in the specification of the "layout" parameter?

pyrigi/graph.py
MatteoGallet: Ok,  I will try something, if it becomes too complicated, I will leave it as it is.

pyrigi/graph.py
Legersky: Sorry for the confusing description: one can specify directly a `placement` (dict from vertices to points). If this parameter is omitted, the method generates a placement using method `layout` and in this parameter, one can specify which layout should be used (`spring`, `circular`, `random` or `planar`).

MatteoGallet (approved):

MatteoGallet (approved):

PR #82, opened 2024-08-27
Feature pebble game algorithm

mihalykoandras:
This adds all features to support the pebble game algorithm that is necessary for the 2d combinatorial rigidity. 

I added a new class directed_graph, that supports the directed graph structure necessary for the pebble game algorithm. I connected the relevant functions of Graph to the directed_graph class. 
Some unit tests are added for both classes.

This is not necessarily a complete implementation of every potential of the pebble game algorithm, but a first - simple - approach for it, that can be improved later down the line. I'm not even 100% sure on the O(|V|^2) nature of this implementation, since that depends on the internal implementations of python. 
But I believe that this is a usable implementation that can be a good basis of further development.


MatteoGallet:
Thanks a lot @mihalykoandras !! I took the liberty to make a couple of minimal changes to that the code would pass the `flake8` check. Is it ok to comment out the two `print` statements in the `is_rigid` method of `Graph` so that the tests will not fail?

mihalykoandras:
Thanks for the flake fix. I removed the unnecessary print() statements.

MatteoGallet:
> Thanks for the flake fix. I removed the unnecessary print() statements.

Thanks a lot!

MatteoGallet:
Would it be ok to transform the comments to the methods at the end of `directed_graph.py` into docstrings? I can take care of that if you agree.

mihalykoandras:
I updated the docstrings. While I believe there are still some unit tests that I'd need to add, I will move it to ready state from draft.

pyrigi/directed_graph.py
```diff
@@ -0,0 +1,313 @@
...
+        """
+        if self.has_node(node_to):
+            tail = edge[0]
+            head = edge[1]
+            self.remove_edge(tail, head)
+            self.add_edge(head, node_to)
+
+    def added_edge_between(self, u: Vertex, v: Vertex) -> {bool, set}:
+        """
+        Checks if edge can be added between the vertices u and v
```
MatteoGallet, line 145: "Checks" -> "Check"

pyrigi/directed_graph.py
```diff
@@ -0,0 +1,313 @@
...
+        It returns if the given edge can be added
```
MatteoGallet, line 146: "It returns if" -> "Return whether"

pyrigi/directed_graph.py
```diff
@@ -0,0 +1,313 @@
...
+        ----------
+        u, v: vertices to add edge between.
+        If u or v is not present in the graph, Error is raised.
+        """
+
+        def dfs(
+            node: Vertex, visited: set, edge_path: list[Edge], current_edge=None
+        ) -> {bool, set}:
+            """
+            Running depth first search to find vertices
```
MatteoGallet, line 158: "Running" -> "Run"

pyrigi/directed_graph.py
```diff
@@ -0,0 +1,313 @@
...
+            that can be reached from u or v.
+            Returns if any of these has outdegree < self._K
```
MatteoGallet, line 160: "Returns if" -> "Return whether"

pyrigi/directed_graph.py
```diff
@@ -0,0 +1,313 @@
...
+        """
+        can_add_edge, fundamental_circuit = self.added_edge_between(u, v)
+        if can_add_edge:
+            return {u, v}
+        else:
+            return fundamental_circuit
+
+    def can_add_edge_between_nodes(self, u: Vertex, v: Vertex) -> bool:
+        """
+        Can you add the edge between the nodes u and v,
```
MatteoGallet, line 259: "Can you add" -> "Check whether one can add"

pyrigi/directed_graph.py
```diff
@@ -0,0 +1,313 @@
...
+        u, v: vertices, between an edge is proposed.
+        If u or v is not present in the graph, Error is raised.
+
+        """
+        can_add_edge, fundamental_circuit = self.added_edge_between(u, v)
+        return can_add_edge
+
+    def add_edge_to_maintain_digraph_if_possible(self, u: Vertex, v: Vertex) -> bool:
+        """
+        Function to add the given edge to the directed graph, if possible.
```
MatteoGallet, line 272: "Function to add" -> "Add"

pyrigi/directed_graph.py
```diff
@@ -0,0 +1,313 @@
...
+                self.add_edges_from([(u, v)])
+            else:
+                self.add_edges_from([(v, u)])
+            return True
+        else:  # if not possible to add, just don't add
+            return False
+
+    def add_edges_to_maintain_out_degrees(self, edges: list[Edge]) -> None:
+        """
+        Simple way to add a list of edges to the directed graph
```
MatteoGallet, line 303: "Simple way to add" -> "Add"

pyrigi/graph.py
```diff
@@ -414,26 +422,123 @@ def max_degree(self) -> int:
         return max([self.degree(v) for v in self.nodes])
 
     @doc_category("Sparseness")
-    def is_sparse(self, K: int, L: int) -> bool:
+    def _build_directed_graph_from_scratch(self, K: int, L: int) -> None:
+        r"""
+        Builds and saves the directed representation of the graph from scratch.
```
MatteoGallet, line 427: "Builds and saves" -> "Build and save"

pyrigi/graph.py
```diff
@@ -414,26 +422,123 @@ def max_degree(self) -> int:
...
+            dir_graph.add_edge_to_maintain_digraph_if_possible(u, v)
+        self.__directed_pebble_graph__ = dir_graph
+
+    # @doc_category("Sparseness")
+    def is_independent(
+        self, u, v, K: int, L: int, use_precomputed_directed_graph: bool = False
+    ) -> bool:
+        r"""
+        Is the (not yet existing) edge between u and v would be
+        :prf:ref:`(K, L)-independent <def-kl-sparse-tight>` from the graph?
```
MatteoGallet, line 452: "Is the (not yet existing) edge between u and v would be :prf:ref:`(K, L)-independent <def-kl-sparse-tight>` from the graph?" -> "Check whether the (not yet existing) edge between u and v would be :prf:ref:`(K, L)-independent <def-kl-sparse-tight>` from the graph."

pyrigi/graph.py
```diff
@@ -414,26 +422,123 @@ def max_degree(self) -> int:
...
+        return self.__directed_pebble_graph__.can_add_edge_between_nodes(u, v)
+
+    # @doc_category("Sparseness")
+    def get_matroid_circuit(
+        self, u, v, K: int, L: int, use_precomputed_directed_graph: bool = False
+    ) -> set:
+        r"""
+        What is the fundamental circuit of the (not yet existing) edge between
+        u and v? If uv is :prf:ref:`(K, L)-independent <def-kl-sparse-tight>`
+        from the graph, it is uv.
```
MatteoGallet, line 471: "What is the fundamental circuit of the (not yet existing) edge between u and v? If uv is :prf:ref:`(K, L)-independent <def-kl-sparse-tight>` from the graph, it is uv." -> "Return the fundamental circuit of the (not yet existing) edge between u and v." I would put the other sentence in the Notes

pyrigi/graph.py
```diff
@@ -414,26 +422,123 @@ def max_degree(self) -> int:
...
+            self._build_directed_graph_from_scratch(K, L)
+
+        return self.__directed_pebble_graph__.to_undirected()
+
+    @doc_category("Sparseness")
+    def _is_directed_graph_sparse(
+        self, K: int, L: int, use_precomputed_directed_graph: bool = False
+    ) -> bool:
+        """
+        Checks if the given directed graph contains exactly the same number of edges
```
MatteoGallet, line 501: "Checks" -> "Check"

MatteoGallet (changes requested):
Requested changes for PEP 257 compliance.

pyrigi/directed_graph.py
```diff
@@ -0,0 +1,313 @@
...
+
+# from pyrigi.misc import doc_category, generate_category_tables
+# from pyrigi.exception import LoopError
+
+"""
+Auxilary class for directed graph used in pebble game style algorithms.
+"""
+
+
+class MultiDiGraph(nx.MultiDiGraph):
```
Legersky, line 13: I think we should use a different name than `MultiDiGraph` since this class is really specific for pebble games as one has to specify K and L. Namely, this is not a class that a user would use for representing a directed multigraph for some other reason than running pebble games. Actually, should this class be exposed to users at all?

pyrigi/directed_graph.py
```diff
@@ -0,0 +1,313 @@
...
+        """
+        return self.__L
+
+    def get_number_of_edges(self) -> int:
+        """
+        Number of directed edges
+        """
+        return len(super().edges)
+
+    def in_degree(self, node: Vertex) -> int:
```
Legersky, line 109: Is there a reason for overriding the parent method?

pyrigi/directed_graph.py
Legersky, line 1: Docstrings of public methods should be formatted according: https://pyrigi.github.io/PyRigi/development/howto.html#docstrings
There has to be an empty line between sections, e.g. above Parameters, to be correctly compiled by Sphinx .

pyrigi/graph.py
```diff
@@ -84,6 +85,13 @@ class Graph(nx.Graph):
     - :doc:`Converting to and from other Data Formats <networkx:reference/convert>`
     """
 
+    """
+    Private variable to save the auxiliary directed graph used for
+    pebble game algorithm here. We don't need to recalculate it necessarily
+    every time.
+    """
+    __directed_pebble_graph__ = pyrigi.directed_graph.MultiDiGraph()
```
Legersky, line 93: Should this be moved to `__init__`? Moreover, it should not be a dunder name, just `_directed_pebble_graph`.

pyrigi/graph.py
```diff
@@ -414,26 +422,123 @@ def max_degree(self) -> int:
...
+
+        dir_graph = pyrigi.directed_graph.MultiDiGraph(K, L)
+        dir_graph.add_nodes_from(self.nodes())
+        for edge in self.edge_list():
+            u, v = edge[0], edge[1]
+            dir_graph.add_edge_to_maintain_digraph_if_possible(u, v)
+        self.__directed_pebble_graph__ = dir_graph
+
+    # @doc_category("Sparseness")
+    def is_independent(
```
Legersky, line 447: I think the name is too general. (Also `get_matroid_circuit`)

pyrigi/graph.py
```diff
@@ -414,26 +422,123 @@ def max_degree(self) -> int:
...
+            not use_precomputed_directed_graph
+            or K != self.__directed_pebble_graph__.get_K()
+            or L != self.__directed_pebble_graph__.get_L()
+        ):
+            self._build_directed_graph_from_scratch(K, L)
+
+        return self.__directed_pebble_graph__.fundamental_circuit(u, v, K, L)
+
+    @doc_category("Sparseness")
+    def get_one_spanning_sparse_subgraph(
```
Legersky, line 484: -> `spanning_sparse_subgraph` ?
Could you add a docstring?

pyrigi/graph.py
```diff
@@ -414,26 +422,123 @@ def max_degree(self) -> int:
...
+            self._build_directed_graph_from_scratch(K, L)
+
+        # all edges are in fact inside the directed graph
+        return (
+            len(self.edge_list())
+            == self.__directed_pebble_graph__.get_number_of_edges()
+        )
+
+    @doc_category("Sparseness")
+    def is_sparse(self, K: int, L: int, combinatorial: bool = True) -> bool:
```
Legersky, line 518: I suggest to switch from `combinatorial` to `algorithm` which would take `"pebble"` or `"subgraph"`. That is more descriptive since both approaches are in some sense combinatorial. (We would like to adopt this also in other parts of the code where we use `combinatorial` now)

pyrigi/graph.py
```diff
@@ -414,26 +422,123 @@ def max_degree(self) -> int:
...
-        for j in range(K, self.number_of_nodes() + 1):
-            for vertex_set in combinations(self.nodes, j):
-                G = self.subgraph(vertex_set)
-                if G.number_of_edges() > K * G.number_of_nodes() - L:
-                    return False
-        return True
+        if combinatorial:
+            return self._is_directed_graph_sparse(K, L)
+
+        else:
```
Legersky, line 532: This should be a fallback option also in the case when pebble games cannot be used because of the restrictions on K and L.

Legersky (commented):
I have added some comments regarding style and names, I will have a look at the algorithm and tests soon.

pyrigi/directed_graph.py
mihalykoandras: Ok

pyrigi/directed_graph.py
mihalykoandras: Ok

pyrigi/directed_graph.py
mihalykoandras: Ok

pyrigi/directed_graph.py
mihalykoandras: Np

pyrigi/graph.py
mihalykoandras: Ok

pyrigi/graph.py
mihalykoandras: Changed

pyrigi/graph.py
mihalykoandras: I changed the description. What is the Notes?

pyrigi/directed_graph.py
```diff
@@ -0,0 +1,313 @@
...
+        # We allow not defining them yet
+        if K is not None and L is not None:
+            self.__check_K_and_L(K, L)
+
+        self.__K = K
+        self.__L = L
+
+        super().__init__(*args, **kwargs)
+
+    def __check_K_and_L(self, K: int, L: int) -> None:
```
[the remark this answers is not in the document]
mihalykoandras: No, it was just a mistake.

pyrigi/directed_graph.py
```diff
@@ -0,0 +1,313 @@
...
+        if 0 >= K:
+            raise ValueError("K must be positive")
+
+        if 0 > L:
+            raise ValueError("L must be non-negative")
+
+        if L >= 2 * K:
+            raise ValueError("L<2K must hold")
+
+    def set_K(self, K: int) -> None:
```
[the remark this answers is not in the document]
mihalykoandras: Like this? I'm not that confident with python syntax as you can see.

pyrigi/directed_graph.py
mihalykoandras: We can rename it, as you wish. What do you suggest? PebbleGraphDirectedGraph?
Probably for the average user it is not necessary to expose this class. I remember to have asked it at the discussion and we decided that for the beginning it can be exposed, and we hide it if needed. How should I hide it?

pyrigi/directed_graph.py
mihalykoandras: Yes. In case we change to hypergraphs, this parent method won't be available any more and I wanted to introduce the framework by than. I would keep it, but I can also remove it, if you prefered like that.

pyrigi/directed_graph.py
mihalykoandras: I changed some, hopefully it is better now. I might have still missed some though

pyrigi/graph.py
mihalykoandras: Is there an __init__ though where I can put it? I don't really think that I should be the one creating the init for graph class.

pyrigi/graph.py
mihalykoandras: What would you suggest instead? For this one and for the get_matroid_circuit?

pyrigi/graph.py
mihalykoandras: I wouldn't change it to spanning_sparse_subgraph, if that is what you suggest. Because there are many different spanning sparse subgraphs of a graph (even maximal ones, like the one present here). So I would like to denote that this returns only one of the possible ones. 

I added the docstrings

pyrigi/graph.py
mihalykoandras: Sure thing. With the fallback option being "pebble" ?

mihalykoandras: I ended up raising an error in case it is different from "pebble" and "subgraph". I hope it works

pyrigi/graph.py
mihalykoandras: Ok, I did that. For that I introduced a new K,L check method, too.

pyrigi/graph.py
MatteoGallet: Thanks! The Notes are an item in the docstring (as, for example, Parameters) whose content should relate to implementation details. In this sense, I believe that the sentence  If uv is `:prf:ref:(K, L)-independent <def-kl-sparse-tight>` from the graph, it is uv." would fit better into that section.

pyrigi/graph.py
```diff
@@ -1029,52 +1171,98 @@ def is_globally_rigid(self, dim: int = 2) -> bool:
...
 
-    @doc_category("Waiting for implementation")
+    @doc_category("Partially implemented")
     def is_Rd_dependent(self, dim: int = 2) -> bool:
         """
         Notes
         -----
          * dim=1: Graphic Matroid
          * dim=2: not (2,3)-sparse
          * dim>=1: Compute the rank of the rigidity matrix and compare with edge count
```
MatteoGallet, line 1181: Please add a blank line before TODO

pyrigi/graph.py
```diff
@@ -1029,52 +1171,98 @@ def is_globally_rigid(self, dim: int = 2) -> bool:
...
+         TODO
```
MatteoGallet, line 1182: Please change 

```
TODO
```

into 

```
TODO
----
```

pyrigi/graph.py
```diff
@@ -1029,52 +1171,98 @@ def is_globally_rigid(self, dim: int = 2) -> bool:
...
+    @doc_category("Partially implemented")
     def is_Rd_independent(self, dim: int = 2) -> bool:
         """
         Notes
         -----
          * dim=1: Graphic Matroid
          * dim=2: (2,3)-sparse
          * dim>=1: Compute the rank of the rigidity matrix and compare with edge count
+
+         TODO
```
MatteoGallet, line 1196: Please change 

```
TODO
```

into 

```
TODO
----
```

pyrigi/graph.py
```diff
@@ -1029,52 +1171,98 @@ def is_globally_rigid(self, dim: int = 2) -> bool:
...
         -----
          * dim=1: Graphic Matroid
-         * dim=2: Remove any edge and it becomes sparse
-           (sparsity for every subgraph except whole graph?)
+         * dim=2: It is not sparse, but remove any edge and it becomes sparse
+                  Fundamental circuit is the whole graph
+         * Not combinatorially:
          * dim>=1: Dependent + Remove every edge and compute the rigidity matrix' rank
+
+         TODO
```
MatteoGallet, line 1224: Please change 

```
TODO
```

into 

```
TODO
----
```

MatteoGallet (changes requested):

pyrigi/graph.py
Legersky: To me, the name sounds like whether the graph is independent, while it is about a (non-)edge. Also it is not clear in what matroid. People might confuse it with the existing `Graph.is_Rd_independent`. 
What about `edge_is_kl_independent`? @MatteoGallet ?
We should also add this notion to the definition `def-kl-sparse-tight`.

pyrigi/graph.py
Legersky: For me,  `spanning_sparse_subgraph` indicates clearly enough what is returned. The fact that it is not unique is immediate from the docstring, see https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.tree.mst.minimum_spanning_tree.html for comparison.

pyrigi/graph.py
mihalykoandras: Ok

pyrigi/graph.py
mihalykoandras: Ok

MatteoGallet (approved):

pyrigi/graph.py
MatteoGallet: I agree with the proposed version `edge_is_kl_independent` !

pyrigi/graph.py
mihalykoandras: Added

pyrigi/directed_graph.py
mihalykoandras: Sure, renamed

pyrigi/directed_graph.py
MatteoGallet: I agree about renaming the class, I think that for example both `PebbleDirectedGraph` and `PebbleDiGraph` would be fine! I think that we may hide this class.

pyrigi/graph.py
mihalykoandras: I'm waiting for your conclusion and change accordingly.

pyrigi/directed_graph.py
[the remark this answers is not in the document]
mihalykoandras: Sorry, I'm not sure I understand it now. Do you prefer me keeping it like this (after the changes) or do you want me to change to @property?

pyrigi/graph.py
mihalykoandras: Renamed to `edge_is_kl_independent`

pyrigi/graph.py
mihalykoandras: Ok, changed.

pyrigi/directed_graph.py
mihalykoandras: Ok, I changed it to PebbleDiGraph and _pebble_di_graph.py .

pyrigi/directed_graph.py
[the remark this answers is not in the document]
Legersky: I meant that `_L` is definitely better than `__L`.
Changing to properties to avoid setters and getters would be nice as well but we can refactor it later if you don't want to bother with it now.

pyrigi/directed_graph.py
Legersky: `_pebble_di_graph.py` or `_pebble_digraph.py`?

pyrigi/graph.py
Legersky: I see, I did not realize that we did not create `__init__` yet as it had not been needed. It really has to go to `__init__`, otherwise it is a class attribute (i.e. shared among all instances of the class), not an instance attribute.
I would say you can create one, something like:
```py
def __init__(self, graph_data=None, **attr):
    super().__init__(incoming_graph_data=graph_data, **attr)
```

pyrigi/graph.py
Legersky: `get_matroid_circuit` to `edge_kl_circuit`?

pyrigi/graph.py
Legersky: Actually, all tests pass even with `_directed_pebble_graph = pyrigi._pebble_di_graph.PebbleDiGraph()` deleted. I think it is because `_build_directed_graph_from_scratch` creates it anyway.

pyrigi/framework.py
```diff
@@ -1081,7 +1081,7 @@ def is_min_inf_rigid(self) -> bool:
         return True
 
     @doc_category("Waiting for implementation")
-    def is_independent(self) -> bool:
+    def edge_is_kl_independent(self) -> bool:
```
MatteoGallet, line 1084: Please revert this change, I believe the method `is_independent` was changed inadvertently.

pyrigi/graph.py
ggmath: I am fine with a "default" parameter for algorithm .

pyrigi/graph.py
mihalykoandras: I'm not 100% sure now, if I can delete it. I want to save the variable for out of the scope, too. Will it be able to reuse it if I remove the `_directed_pebble_graph = pyrigi._pebble_di_graph.PebbleDiGraph() `

pyrigi/graph.py
mihalykoandras: Done

pyrigi/framework.py
mihalykoandras: Correct, thanks.

pyrigi/graph.py
mihalykoandras: Let it be the 'default' path then.

pyrigi/graph.py
Legersky: Yes, I think so.

mihalykoandras: Sure, removed it.

pyrigi/directed_graph.py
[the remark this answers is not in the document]
mihalykoandras: I changed it to properties. This is the very first time I use properties, so I might not get the idea at all. So please check this, if this is what you expected. I put it to the very end so that it can be reversed easily.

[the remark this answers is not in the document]
Legersky: I looks good, thanks.

test/test_pebble_di_graph.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+def test_out_degree():
+    graph = OutStar(2, 1, 1)
+    assert graph.out_degree(0) == 2
+    assert graph.out_degree(1) == 0
+    assert graph.out_degree(2) == 0
+
+
+def test_redirect_edge():
```
Legersky, line 97: Should this be part of `test_point_edge_head_to` instead?

test/test_pebble_di_graph.py
```diff
@@ -0,0 +1,168 @@
...
+    assert 0 == Cycle_graph.out_degree(0)
+    assert 2 == Cycle_graph.out_degree(1)
+
+
+def test_point_edge_head_to():
+    graph = DirectedPath(3, 2, 2)
+    graph.point_edge_head_to((0, 1), 2)
+    assert (0, 1) not in graph.edges
+    assert (1, 2) in graph.edges
+    assert (1, 2) in graph.edges
```
Legersky, line 110: `(1,0)` instead?

pyrigi/graph.py
```diff
@@ -566,27 +575,47 @@ def is_sparse(self, K: int, L: int, algorithm: str = "pebble") -> bool:
...
+            else:
+                # otherwise use "subgraph"
+                algorithm = "subgraph"
+            return self.is_sparse(K, L, algorithm)
+
+        # reaching this position means that the algorithm is unknown
+        raise ValueError(
+            'The algorithm needs to be either "pebble",\
+             "subgraph", "default" or not specified.'
+        )
```
MatteoGallet, line 602: In order to be a bit more informative, this could be something like:

```
        raise ValueError(
            f'If specified, the value of the algorithm parameter must be one of 
            "pebble", "subgraph", or "default". Insted, it is {algorithm}.'
        )
```

pyrigi/graph.py
```diff
@@ -88,6 +89,12 @@ class Graph(nx.Graph):
     - :doc:`Converting to and from other Data Formats <networkx:reference/convert>`
     """
 
+    """
```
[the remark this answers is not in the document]
mihalykoandras: Done

pyrigi/graph.py
```diff
@@ -417,36 +424,204 @@ def max_degree(self) -> int:
...
+
+        dir_graph = pyrigi._pebble_di_graph.PebbleDiGraph(K, L)
+        dir_graph.add_nodes_from(self.nodes())
+        for edge in self.edge_list():
+            u, v = edge[0], edge[1]
+            dir_graph.add_edge_to_maintain_digraph_if_possible(u, v)
+        self._directed_pebble_graph = dir_graph
+
+    # @doc_category("Sparseness")
+    def edge_is_kl_independent(
```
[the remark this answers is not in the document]
mihalykoandras: I'm not familiar of how to do that. Would you please add it to the definitions?

pyrigi/graph.py
```diff
@@ -417,36 +424,204 @@ def max_degree(self) -> int:
...
+        self, u, v, K: int, L: int, use_precomputed_directed_graph: bool = False
+    ) -> bool:
+        r"""
+        Check whether the (not yet existing) edge between u and v
+        would be :prf:ref:`(K, L)-independent <def-kl-sparse-tight>` from the graph.
+        """
+
+        if (
```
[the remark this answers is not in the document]
mihalykoandras: Yes, I believe that is caused by removing the `    _directed_pebble_graph = pyrigi._pebble_di_graph.PebbleDiGraph()
` line. With that it should work. But yes, I'll place a warning.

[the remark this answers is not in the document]
mihalykoandras: Also, I see that the current `is_Rd_circuit` might also need a similar parameter

pyrigi/graph.py
```diff
@@ -417,36 +424,204 @@ def max_degree(self) -> int:
...
+            self._build_directed_graph_from_scratch(K, L)
+
+        return self._directed_pebble_graph.can_add_edge_between_nodes(u, v)
+
+    # @doc_category("Sparseness")
+    def edge_kl_circuit(
+        self, u, v, K: int, L: int, use_precomputed_directed_graph: bool = False
+    ) -> set:
+        r"""
+        Return the fundamental circuit of the (not yet existing) edge between u and v.
```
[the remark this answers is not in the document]
mihalykoandras: Again, I'm not familiar with the working of the definitions. If you could explain what is expected, I could add those definitions. But probably it is easier if you add them.

pyrigi/graph.py
```diff
@@ -417,36 +424,204 @@ def max_degree(self) -> int:
...
 
+        Parameters
+        ----------
+        K:
+        L:
+        algorithm:
+            "pebble" or "subgraph".
+            If "pebble", the function uses the pebble game algorithm to check
+            for sparseness. If "subgraph", it uses the subgraph method.
+            If not specified, it defaults to "pebble".
```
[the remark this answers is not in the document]
mihalykoandras: Updated.

pyrigi/graph.py
```diff
@@ -417,36 +424,204 @@ def max_degree(self) -> int:
...
+        # reaching this position means that the algorithm is unknown
+        raise ValueError(
+            'The algorithm needs to be either "pebble",\
+             "subgraph", "default" or not specified.'
+        )
 
     @doc_category("Sparseness")
-    def is_tight(self, K: int, L: int) -> bool:
+    def is_tight(self, K: int, L: int, algorithm: str = "pebble") -> bool:
         r"""
```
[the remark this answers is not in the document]
mihalykoandras: Sure, I changed them

pyrigi/graph.py
```diff
@@ -1036,52 +1209,102 @@ def is_globally_rigid(self, dim: int = 2) -> bool:
...
+                    return False
+            return True
+
+        if dim == 2:
+            # get max sparse sugraph and check the fundamental circuit of
+            # the one last edge
+            if self.number_of_edges() != 2 * self.number_of_nodes() - 2:
+                return False
+            max_sparse_subgraph = self.spanning_sparse_subgraph(
+                K=2, L=3, use_precomputed_directed_graph=True
```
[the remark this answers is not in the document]
mihalykoandras: Yes, I updated. There might be a need for `use_precomputed_directed_graph` option though

pyrigi/matroid.py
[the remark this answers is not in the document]
mihalykoandras: True, sorry for that. Removed

test/test_matroid.py
```diff
@@ -19,7 +19,7 @@ def test_linear_matroid_rank():
 def test_linear_matroid_independence():
     A = Matrix([[1, 2, 3], [2, 4, 6], [3, 4, 5]])
     M = LinearMatroid(A)
-    assert M.is_independent([0, 2])
+    assert M.edge_is_kl_independent([0, 2])
```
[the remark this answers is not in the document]
mihalykoandras: Done

test/test_pebble_di_graph.py
mihalykoandras: Correct, this is a remainder of an earlier design.

test/test_pebble_di_graph.py
mihalykoandras: In fact the problem ran much deeper. We should not be able to redirect an edge to an external vertex. This should be a reversing function, only implemented like this, because later with hypergraphs this is the preferred way. I hopefully fixed it now.

pyrigi/graph.py
mihalykoandras: Changed so.

pyrigi/graph.py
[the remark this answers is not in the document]
MatteoGallet: Sure, I can take care of that. Could you please point me to the precise mathematical definition that you implemented in the algorithm? Thanks!

pyrigi/graph.py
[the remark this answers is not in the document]
MatteoGallet: Sure, I can take care of that as well. Could you please point me to the precise mathematical definition that you implemented in the algorithm? Thanks!

pyrigi/graph.py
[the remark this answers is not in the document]
mihalykoandras: https://arxiv.org/pdf/2209.06204 p2

pyrigi/graph.py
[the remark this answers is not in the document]
mihalykoandras: https://egres.elte.hu/tr/egres-14-12.pdf
3.5 Section, p51. It is a bit convoluted.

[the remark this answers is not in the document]
mihalykoandras: What is a circle in the (1,1)-count matroid (graphical matroid)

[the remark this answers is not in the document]
MatteoGallet: Thanks!

pyrigi/graph.py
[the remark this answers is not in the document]
MatteoGallet: Thanks!

pyrigi/directed_graph.py
mihalykoandras: Can you confirm that this naming is acceptable?

pyrigi/graph.py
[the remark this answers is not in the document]
Legersky: yes, I agree.

pyrigi/directed_graph.py
MatteoGallet: Hi! sorry for the delay! I would say that `_pebble_digraph` may be slightly better than `_pebble_di_graph` since the word "digraph" is attested, for example in [Wikipedia](https://en.wikipedia.org/wiki/Directed_graph).

pyrigi/_pebble_di_graph.py
```diff
@@ -0,0 +1,343 @@
...
+                edge_path.pop()
+            return False, visited
+
+        max_degree_for_u_and_v_together = 2 * self.K - self.L - 1
+
+        if not self.has_node(u):
+            raise ValueError(
+                "Cannot check if edge can be added, since Vertex "
+                + u
+                + " is not present in graph."
```
Legersky, line 221: Could you replace this (and also below) by
```
f"Vertex {u} is not present in the graph."
```

pyrigi/_pebble_di_graph.py
```diff
@@ -0,0 +1,343 @@
...
+        Number of edges leading out from a node.
+
+        Parameters
+        ----------
+        node: Vertex, that we wish to know the outdegree.
+        TODO check if vertex exists
+        """
+        return super().out_degree(node)
+
+    def redirect_edge_to_head(self, edge: Edge, node_to: Vertex) -> None:
```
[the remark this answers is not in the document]
Legersky: I have fixed `Edge` and created `DirectedEdge`, see #95.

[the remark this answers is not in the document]
mihalykoandras: Yes, this is preparation for the hypergraph solution, when reversing an edge doesn't make any sense, but claiming a point to be the head now makes sense. It might be a bit confusing now, but I would keep it unless you insist on the change.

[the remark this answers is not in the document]
MatteoGallet: I see, thanks for the explanation! I would say, then please leave it as it is.

pyrigi/directed_graph.py
mihalykoandras: Sure, done

pyrigi/_pebble_di_graph.py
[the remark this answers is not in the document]
mihalykoandras: I'm unsure now, what do you mean by fixing? So what you suggest me to do is to use DirectedEdge in everywhere in the _pebble_digraph class?

pyrigi/_pebble_di_graph.py
mihalykoandras: Sure, fixed.

pyrigi/_pebble_di_graph.py
```diff
@@ -0,0 +1,343 @@
...
+        and the fundamental (matroid) cycle of the edge uv.
+
+        Parameters
+        ----------
+        u, v: vertices to add edge between.
+        If u or v is not present in the graph, Error is raised.
+        """
+
+        def dfs(
+            node: Vertex, visited: set, edge_path: list[Edge], current_edge=None
```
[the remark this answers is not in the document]
mihalykoandras: Fixed

pyrigi/_pebble_di_graph.py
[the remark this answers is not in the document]
mihalykoandras: Ok, I did just that, please check if you wished like this.

pyrigi/graph.py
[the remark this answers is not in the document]
mihalykoandras: Ok, I added  use_precomputed_pebble_digraph to hopefully all relevant places.

pyrigi/graph.py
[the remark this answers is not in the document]
mihalykoandras: I placed use_precomputed_pebble_digraph hopefully in every relevant place now and added comments in every function, with always being the default value False.

pyrigi/graph.py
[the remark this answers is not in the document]
Legersky: Thanks, I think so. I would just change to warning a bit: I guess it is not only about `AtributeError`, but the result might be just wrong if the graph was modified without appropriate change of the pebble digraph, cannot it?

pyrigi/_pebble_di_graph.py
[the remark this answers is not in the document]
Legersky: Thanks, that's what I meant. Now it is clear from the signature that the order of vertices matters.

pyrigi/graph.py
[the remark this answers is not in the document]
mihalykoandras: So what should I change it to?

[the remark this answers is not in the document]
Legersky: What about 'Use True only if you are certain that the pebble game directed graph is consistent with the graph.'?
I can change it.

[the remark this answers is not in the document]
MatteoGallet: Thanks! We'll have a look at the docstrings and then we will merge the branch if we have no further questions.

Legersky (approved):

PR #85, opened 2024-09-19
Bug in is_rigid for d=1 fixed

Legersky:


ggmath:
Maybe we should check is_min_rigid as well

ggmath (approved):

PR #123, opened 2024-11-07
Feature realization number

jcapco:
Feature is there. Probably documentation need some improvement.

### DIFF
--- a/pyrigi/graph.py
+++ b/pyrigi/graph.py
@@ -7,6 +7,7 @@ from __future__ import annotations
 from copy import deepcopy
 from itertools import combinations
 from typing import List, Union, Iterable
+from lnumber import lnumber, lnumbers
 
 import networkx as nx
 import matplotlib.pyplot as plt
@@ -485,7 +486,7 @@ class Graph(nx.Graph):
             self._build_pebble_digraph(K, L)
 
         return self._pebble_digraph.to_undirected()
-
+        
     @doc_category("Sparseness")
     def _is_pebble_digraph_sparse(
         self, K: int, L: int, use_precomputed_pebble_digraph: bool = False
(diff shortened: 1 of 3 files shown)


MatteoGallet (commented):
Thanks for your contribution! There is a linting problem, could you please use 

```
black .
```

on your files?

pyrigi/graph.py
```diff
@@ -972,6 +973,43 @@ def extension_sequence(
...
+        >>> G.number_of_minimally_rigid_realizations(spherical_realizations=True)
+        2
+        """        
+        n = self.to_int()
+        if not self.is_min_rigid():
+          return 0
+        if spherical_realizations :
+          return lnumbers(n)
+        else:
+          return lnumber(n)
```
MatteoGallet, line 1011: Could you please change `minimally` -> `min` ?

MatteoGallet (dismissed):

pyrigi/graph.py
Legersky: I would go just for `number_of_realizations`.

Legersky: And I think we should check that it is minimally rigid by default.

Legersky: If the graph is not minimally rigid, I would throw a `ValueError`.

pyrigi/graph.py
```diff
@@ -972,6 +973,44 @@ def extension_sequence(
...
+
+        Examples
+        --------
+        >>> from pyrigi import Graph
+        >>> G = Graph([(0,1),(1,2),(2,0)])
+        >>> G.number_of_minimally_rigid_realizations() #returns number of planar realizations
+        2
+        >>> G.number_of_minimally_rigid_realizations(spherical_realizations=True)
+        2
+        """
```
MatteoGallet, line 1005: Could you please change this line into:

```
""" #noaq: E501
```

since some line in the doc string exceeds 90 characters and so the linting test fails?

MatteoGallet (approved):

PR #124, opened 2024-11-07
Feature Onshape - generate models for 3D printing

hucik14:
First draft to be reviewed, please @Legersky 

Yet to be done before the pull:
- add tests 
- add docs

### DIFF
--- a/pyrigi/framework.py
+++ b/pyrigi/framework.py
@@ -1324,6 +1324,11 @@ class Framework(object):
         Return the lengths (numerically) of all edges in the framework.
 
         The ordering is given by graph().edge_list() method.
+
+        Returns
+        -------
+        pair_lengths
+            List of lengths of the edges in the framework.
         """
         from numpy import array as nparray
         from numpy.linalg import norm as npnorm
--- a/pyrigi/framework.py
+++ b/pyrigi/framework.py
@@ -1345,7 +1345,7 @@ class Framework(object):
     @doc_category("Other")
     def generate_onshape_parameters_for_3d_print(self,
                                                  scale: float = 1.0,
-                                                 roundings: int = 3) -> list:
+                                                 roundings: int = 3) -> tuple:
         """
         Generate OnShape CAD details for models.
 
@@ -1360,23 +1360,30 @@ class Framework(object):
 
         Returns
         -------
+        onshape_bars_gen_url
+            String URL to the OnShape model.
         readable_form
             List of scaled and rounded lengths of the edges in the framework.
 
-        """
-        print('Copy the following model to your OnShape workspace:')
-        print('https://cad.onshape.com/documents/6b5c6a508178ccdc56722495/w/5477a320ec050694840763d5/e/4246fa25bf9c77c9dd0d0fe2?renderMode=0&uiState=672cc055af05270e60941bce')
-        print('Use the following array of lengths as input for Bars feature:')
+        Examples
+        --------
+        >>> F = Framework(Graph([(0,1), (1,2), (2,3), (0,3)]), {0:[0,0], 1:[1,0], 2:[1,'1/2 * sqrt(5)'], 3:[1/2,'4/3']})
+        >>> url, l = F.generate_onshape_parameters_for_3d_print(scale=10, roundings=2)
+        >>> print(url)
+        >>> print(l)
 
-        edges_lengths = self.get_edges_lengths()
+        """
+        onshape_bars_gen_url = ('https://cad.onshape.com/documents/6b5c6a508178ccdc'
+                                '56722495/w/5477a320ec050694840763d5/e/4246fa25bf9c'
+                                '77c9dd0d0fe2')
+
+        edges_lengths = self.edge_lengths()
+
+        # round and scale the lengths of the edges
         readable_form = [float(round(scale * length, roundings))
                          for length in edges_lengths]
 
-        # round and convert to string
-        print('Array of distances:')
-        print(readable_form)
-
-        return readable_form
+        return onshape_bars_gen_url, readable_form
 
 
 Framework.__doc__ = Framework.__doc__.replace(
--- a/pyrigi/framework.py
+++ b/pyrigi/framework.py
@@ -1319,9 +1319,9 @@ class Framework(object):
         return new_framework
 
     @doc_category("Other")
-    def generate_onshape_script(self):
+    def get_edges_lengths(self) -> list:
         """
-        Generate OnShape CAD Feature Script for models
+        Return the lengths (numerically) of all edges in the framework.
         """
         from numpy import array as nparray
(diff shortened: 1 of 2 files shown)


pyrigi/framework.py
```diff
@@ -1318,6 +1318,61 @@ def rotate2D(self, angle: float, inplace: bool = True) -> Union[None, Framework]
         new_framework.rotate2D(angle, True)
         return new_framework
 
+    @doc_category("Other")
+    def get_edges_lengths(self) -> list:
```
hucik14, line 1322: This method calculates the lengths, I used numpy for simplification, but maybe it should be done different way more concise with the package

pyrigi/framework.py
```diff
@@ -1318,6 +1318,66 @@ def rotate2D(self, angle: float, inplace: bool = True) -> Union[None, Framework]
...
+        """
+        from numpy import array as nparray
+        from numpy.linalg import norm as npnorm
+
+        points = self.realization(as_points=True)
+        pairs = self.graph().edge_list()
+        pair_lengths = [npnorm(nparray(points[pair[0]],
+                                       dtype='float64') - nparray(points[pair[1]],
+                                                                  dtype='float64'))
+                        for pair in pairs]
```
MatteoGallet, line 1341: Could we use `self._graph.edges` rather than `self.graph().edge_list()` ? 
Moreover, I think it would be nice if the output is a dictionary whose keys are the edges, so that one does not have to rely on a specific ordering.

pyrigi/framework.py
```diff
@@ -1318,6 +1318,66 @@ def rotate2D(self, angle: float, inplace: bool = True) -> Union[None, Framework]
...
+        print('https://cad.onshape.com/documents/6b5c6a508178ccdc56722495/w/5477a320ec050694840763d5/e/4246fa25bf9c77c9dd0d0fe2?renderMode=0&uiState=672cc055af05270e60941bce')
+        print('Use the following array of lengths as input for Bars feature:')
+
+        edges_lengths = self.get_edges_lengths()
+        readable_form = [float(round(scale * length, roundings))
+                         for length in edges_lengths]
+
+        # round and convert to string
+        print('Array of distances:')
+        print(readable_form)
```
MatteoGallet, line 1377: Rather than printing on screen, could the function return the two strings (url and `readable_form`) so that the user can decide how to use them (display/save/etc)?

pyrigi/framework.py
Legersky: We should have such method in general. You can keep it numeric and we refactor it later to provide symbolic output depending on a parameter. But I agree with Matteo to return dictionary "edge -> its length" to avoid troubles with the order of the edges. I would also rename it to `edge_lengths`.

pyrigi/framework.py
Legersky: The order of `self._graph.edges` depends on the way how the graph was constructed!

hucik14: So which one of these two is more preferable?

Legersky: I'd say the dictionary. The only problem here is how to sort the vertices of an edge in case they are not comparable.

pyrigi/framework.py
hucik14: I added this TODO in the description and renamed it > resolved

pyrigi/framework.py
hucik14: done

pyrigi/framework.py
```diff
@@ -1318,6 +1318,75 @@ def rotate2D(self, angle: float, inplace: bool = True) -> Union[None, Framework]
...
+            String URL to the OnShape model.
+        readable_form
+            List of scaled and rounded lengths of the edges in the framework.
+
+        Examples
+        --------
+        >>> F = Framework(Graph([(0,1), (1,2), (2,3), (0,3)]), {0:[0,0], 1:[1,0], 2:[1,'1/2 * sqrt(5)'], 3:[1/2,'4/3']})
+        >>> url, l = F.generate_onshape_parameters_for_3d_print(scale=10, roundings=2)
+        >>> print(url)
+        >>> print(l)
```
Legersky, line 1376: Here should be also the outputs, otherwise doctesting will fail.

pyrigi/framework.py
```diff
@@ -1318,6 +1318,75 @@ def rotate2D(self, angle: float, inplace: bool = True) -> Union[None, Framework]
...
+
+        """
```
Legersky, line 1378: There are too long lines in the input to pass PEP8 check by `flake8`, please, see https://pyrigi.github.io/PyRigi/development/howto.html#pep8

Legersky (commented):
Do I understand correctly that now it outputs disassembled links, so you do not care about the order of the edges nor the graph?

pyrigi/framework.py
hucik14: yes, for the first version it will output only disassembled links

pyrigi/framework.py
hucik14: now it returns dictionary, I kept `self.graph().edge_list()` but this can be changed, just let me know

pyrigi/framework.py
```diff
@@ -1319,9 +1319,9 @@ def rotate2D(self, angle: float, inplace: bool = True) -> Union[None, Framework]
         return new_framework
 
     @doc_category("Other")
-    def edge_lengths(self) -> list:
+    def edge_lengths(self) -> dict:
```
Legersky, line 1322: -> dict[Edge,float]

hucik14: it is more dict[tuple(list),float]... because edge_list() returns list... but I can keep it as dict[Edge,float] if it is more useful for a user, as you prefer

pyrigi/framework.py
Legersky: `self.graph().edge_list()` creates a copy of the graph, I would do:
```python
return {tuple(pair) : npnorm(nparray(points[pair[0]],
                                   dtype='float64') - nparray(points[pair[1]],
                                                                 dtype='float64'))
                       for pair in self._graph.edges }

```

pyrigi/framework.py
Legersky: image

hucik14: if it doesn't work now because of too long line, I can skip the check for print output

pyrigi/framework.py
Legersky: `Edge` is defined in `data_type.py`, so it should be fine.

pyrigi/framework.py
Legersky: `# noqa: E501` after the docstring ;)

Legersky (approved):

PR #126, opened 2024-11-07
Added Examples and Tests for `rigidity_matrix_rank` (including n-Frustum)

matthiashimmelmann:
Replaces #121


MatteoGallet (commented):
Thanks! Could you please modify the `md` file so that the definition is "purely mathematical", not as a Jupyter notebook, with no direct references to `pyrigi` functions in the text, but only in the dedicated part at the end?

MatteoGallet (approved):

PR #131, opened 2024-11-07
Many new examples and tests added to resolve TODO

matthiashimmelmann:
Alison and Rebecca provided a multitude of examples, some of which I recycled as tests.

### DIFF
--- a/pyrigi/framework.py
+++ b/pyrigi/framework.py
@@ -258,9 +258,11 @@ class Framework(object):
         """
         Return a copy of the underlying graph.
 
-        TODO
+        Examples
         ----
-        example
+        >>> F = Framework.Random(Graph([(0,1), (1,2), (0,2)]))
+        >>> F.graph()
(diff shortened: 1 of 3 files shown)


pyrigi/framework.py
```diff
@@ -903,8 +929,8 @@ def stress_matrix(
                     stress_matr[i, i] += stress[edge_order.index(edge)]
         for v, w in combinations(self._graph.nodes, 2):
             i, j = vertex_list.index(v), vertex_list.index(w)
-            correct_edge = (
-                lambda edge: edge if edge in edge_order else [edge[1], edge[0]]
+            correct_edge = lambda edge: (
+                edge if edge in edge_order else [edge[1], edge[0]]
```
MatteoGallet, line 933: This creates a problem with the linter: 

```
E731 do not assign a lambda expression, use a def
```

MatteoGallet (approved):

PR #136, opened 2024-11-08
Refactor realization number

Legersky:
I have shortened the name of the method.
The definition and test are missing.

### DIFF
--- a/pyrigi/graph.py
+++ b/pyrigi/graph.py
@@ -1001,7 +1001,7 @@ class Graph(nx.Graph):
         return None if return_solution else False
 
     @doc_category("Generic rigidity")
-    def number_of_min_rigid_realizations(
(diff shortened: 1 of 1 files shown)


MatteoGallet (approved):

Legersky:
We have to decide whether we count the reflections and modify the output accordingly if not.
The output from `lnumber` is strange for $K_2$.

jcapco:
The strange number in for K2 (or less number of vertices) is coming from the C code that has to be changed. This is known (first pointed out by Sean Dewar). The implementation is not correct if the number of vertices is less than 3. This will be fixed in the next lnumber release. Meanwhile please output the correct number for these trivial cases.

MatteoGallet (approved):

PR #137, opened 2024-11-08
Refactor frustum

Legersky:
I have improved the linking between the source and the documentation and distributed the tests into existing ones.

Before we merge, we should fix the definition, since it does not determine the graph uniquely: the following one satisfies the definition as well. I think that what you want is the cartesian product of n-cycle and an edge, don't you?

image


matthiashimmelmann:
@Legersky Yes, the definition is obviously flawed. I initially had a more convoluted definition that was shortened to something non-unique. Indeed, the n-Frustum is just the Cartesian product of an n-cycle and an edge, or on a framework level the Minkowski sum of a regular polygon and an edge. Needless to say, that definition is more elegant.

MatteoGallet (approved):

PR #147, opened 2024-11-09
Fixed bug in  trivial_inf_flexes

MatteoGallet:
Bug reported by Sean Dewar: the computation of `inf_flexes` was wrong, and in particular it depended on the order in which the vertices appeared in the construction of the graph.
This PR fixes the bug by making uniform between the methods `rigidity_matrix`, `trivial_inf_flexes`, and `inf_flexes` the choice of vertex and edge order.


Legersky:
Could you add the descriptions of `edge_order` and `vertex_order` to the docstrings? And fix the description of `edge_order` in `rigidity_matrix` as well?
Please, add Sean's example to the tests.

pyrigi/framework.py
```diff
@@ -1019,6 +1034,15 @@ def nontrivial_inf_flexes(
         -----------
         :prf:ref:`Infinitesimal flex <def-inf-rigid-framework>`
 
+        Parameters
+        ----------
+        vertex_order:
+            A list of vertices, providing the ordering for the entries
+            of the infinitesimal flexes.
+        edge_order:
```
Legersky, line 1042: I think in `nontrivial_inf_flexes` the `edge_order` does not make sense.

pyrigi/framework.py
```diff
@@ -1080,6 +1104,12 @@ def inf_flexes(
         include_trivial:
             Boolean that decides, whether the trivial motions should
             be included (``True``) or not (``False``)
+        vertex_order:
+            A list of vertices, providing the ordering for the entries
+            of the infinitesimal flexes.
+        edge_order:
```
Legersky, line 1110: I think here the `edge_order` does not make sense either.

pyrigi/framework.py
```diff
@@ -1115,14 +1158,26 @@ def inf_flexes(
         return basis[s:]
 
     @doc_category("Infinitesimal rigidity")
-    def stresses(self) -> List[Matrix]:
+    def stresses(
+        self,
+        vertex_order: List[Vertex] = None,
```
Legersky, line 1163: Here the `vertex_order` does not make sense.

pyrigi/framework.py
```diff
@@ -1420,14 +1473,255 @@ def inf_flexes(
         return basis[s:]
 
     @doc_category("Infinitesimal rigidity")
-    def stresses(self) -> List[Matrix]:
+    def is_inf_flex(
```
Legersky, line 1476: We already have `is_dict_inf_flex` and `is_vector_inf_flex`, so we should unify this somehow.

matthiashimmelmann: True, I forgot about these. I will unify the methods in the next PR. Also, I'll try to avoid sounding like `is_inf_flexible` by reusing the old names.

pyrigi/framework.py
```diff
@@ -1352,6 +1375,8 @@ def trivial_inf_flexes(self) -> List[Matrix]:
     @doc_category("Infinitesimal rigidity")
     def nontrivial_inf_flexes(
         self,
+        vertex_order: List[Vertex] = None,
+        edge_order: List[Edge] = None,
```
Legersky, line 1379: `edge_order` is not important for inf flexes.

pyrigi/framework.py
```diff
@@ -1387,11 +1419,15 @@ def nontrivial_inf_flexes(
...
-        return self.inf_flexes(include_trivial=False)
+        return self.inf_flexes(
+            vertex_order=vertex_order, edge_order=edge_order, include_trivial=False
+        )
 
     @doc_category("Infinitesimal rigidity")
     def inf_flexes(
         self,
+        vertex_order: List[Vertex] = None,
+        edge_order: List[Edge] = None,
```
Legersky, line 1430: `edge_order` is not important for inf flexes.

pyrigi/framework.py
```diff
@@ -2064,38 +2134,69 @@ def _transform_inf_flex_to_pointwise(  # noqa: C901
...
 
+    @doc_category("Infinitesimal rigidity")
     def is_vector_inf_flex(
-        self, vect: Matrix, vertex_order: List[Vertex] = None
+        self,
+        inf_flex: List[Coordinate],
+        vertex_order: List[Vertex] = None,
+        numerical: bool = False,
+        numerical_digits: int = 35,
+        numerical_atol: float = 1e-10,
```
Legersky, line 2144: Could a reasonable number of digits be just determined from the tolerance?

pyrigi/framework.py
```diff
@@ -2064,38 +2134,69 @@ def _transform_inf_flex_to_pointwise(  # noqa: C901
...
     ) -> bool:
-        """
-        Return whether a vector is an infinitesimal flex of the framework.
+        r"""
+        Tests whether an infinitesimal flex lies in the kernel of the rigidity matrix.
```
Legersky, line 2147: To be honest, I liked the original description more since an infinitesimal flex is in the kernel by definition, while here we want to return whether a given vector is an infinitesimal flex.

Legersky (approved):

PR #149, opened 2024-11-12
Added `C_n`-symmetric frameworks to `frameworkDB`

matthiashimmelmann:
These frameworks pose a counterexample for the symmetry-adjusted Laman count. They have been described by Alison La Porta.


doc/math/graph-theory/known-families.md
```diff
@@ -7,6 +7,44 @@ Here we introduce several known graphs that can be accessed in the PyRigi graph
...
+
+Let $n\geq8$ be even, and let $C_n$ denote the anti-clockwise rotation around the origin by $2\pi/n$. 
+Define a {prf:ref}`framework <def-framework>` $(G,p)$ on $n$ joints $\{p(v_1),\dots,p(v_n)\}$ such that $p(v_1)=C_np(v_n)$ and $p(v_i)=C_np(v_{i-1})$ for all $2\leq i\leq n$. In addition, each joint $p(v)$ of $(G,p)$, $p(v)$ is adjacent exactly to the following joints: $C_np(v)$, $C_n^{-1}p(v)$, $C_n^3p(v)$ and $C_n^{-3}p(v)$. For some vector $t$ on the line from the origin to a joint in $\{v_1,\dots,v_n\}$, such a framework has an {prf:ref}`infinitesimal motion <def-inf-flex>` $m$ which satisfies the system of equations
+\begin{equation*}
+    m(v_i)=\begin{cases}
+        t & \text{if } i \text{ is even}\\
+        -t & \text{if } i \text{ is odd},
+    \end{cases}
+\end{equation*}
+where $1\leq i\leq n$.
```
MatteoGallet, line 28: I would suggest to use "vertex" instead of "joint" and to specify that the realization is in the plane? Moreover, I think it would be better to first specify the edges in G, and not to say that one realization of a vertex is adjacent to other realizations of vertices.

Legersky:
Could you break the lines to 90 chars, please?

MatteoGallet (approved):

PR #152, opened 2024-11-14
Added `Graph` method `max_rigid_dimension`

matthiashimmelmann:
The handles Tony's request about the maximum rigid dimension.

### DIFF
--- a/pyrigi/graph.py
+++ b/pyrigi/graph.py
@@ -1969,14 +1969,14 @@ class Graph(nx.Graph):
         d = 1
         while True:
             # Find the smallest d for which the Maxwell count holds.
-            if E < d*V-(d+1)*d/2 or 2*V == d+1:
+            if E < d * V - (d + 1) * d / 2 or 2 * V == d + 1:
                 dim = d
                 break
-            d = d+1
+            d = d + 1
 
         for d in range(dim, 0, -1):
             if self.is_rigid(d, combinatorial=False):
-                if d == 2*V-1 and self.is_rigid(d, combinatorial=False):
+                if d == 2 * V - 1 and self.is_rigid(d, combinatorial=False):
                     # If the dimension is sufficiently large and the graph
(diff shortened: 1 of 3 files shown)


pyrigi/graph.py
```diff
@@ -1937,6 +1937,45 @@ def min_rigid_subgraphs(self, dim: int = 2) -> List[Graph]:
...
+        3
+        """
+        if nx.number_of_selfloops(self) > 0:
+            raise LoopError()
+
+        V = self.number_of_nodes()
+        E = self.number_of_edges()
+        if E == V * (V - 1) / 2:
+            return math.inf
+        # Find the larger root of the Maxwell function E = d*V-d*(d+1)/2
```
MatteoGallet, line 1971: What about 

```
# Find the largest d such that d*(d+1)/2 - d*V + E = 0
```

?

matthiashimmelmann: I was wondering if we can do `floor` as well here? I guess it doesn't save much, just a small nitpick.

Legersky: Yes, I think we can.
We can also directly output 0 if the graph is disconnected.

Legersky:
I am thinking whether we should use `sympy.oo` instead of `math.inf` to have a "symbolic" infinity rather than float?

pyrigi/graph.py
```diff
@@ -1937,6 +1937,51 @@ def min_rigid_subgraphs(self, dim: int = 2) -> List[Graph]:
                         rigid_subgraphs[H2] = False
         return [list(H) for H, is_min in rigid_subgraphs.items() if is_min]
 
+    @doc_category("Generic rigidity")
+    def max_rigid_dimension(self) -> int | Inf:
+        """
+        Compute the maximum dimension, in which a graph is generically rigid.
```
MatteoGallet, line 1943: Could you please link to the math doc?

MatteoGallet (approved):

PR #157, opened 2024-11-19
Medium sized tests

Legersky:
The goal of this PR is to run tests that take a bit longer only when merging to main.


pyproject.toml
matthiashimmelmann, line 1: Doesn't `medium` ideologically replace `slow`? Should we turn `slow` into `long`? Or should we do that on a case-by-case basis as is done in `realization_counting`?

matthiashimmelmann: Or we do it the other way round, call `medium` `slow` instead and replace long with `unsuitable_for_automatic_testing` or `infeasibly_large` to highlight that these tests are not automatically run?

matthiashimmelmann (commented):
I agree that `medium` is not ideal. Maybe we can be more descriptive here.`tests_for_dev`, `slow_tests_for_main` and `local_tests` would be another option.

Legersky:
`realization_counting` is different in the sense that it tests functionality requiring an optional package.

I like the idea of indicating when they are run, what about the following:
```
markers = [
    "slow_main: marks slower tests run at merge to main",
    "long_local: marks long tests run locally",
    "realization_counting: marks tests using extra realization-counting",
    "meshing: marks tests using extra meshing"
]
```
?

matthiashimmelmann:
> `realization_counting` is different in the sense that it tests functionality requiring an optional package.
> 
> I like the idea of indicating when they are run, what about the following:
> 
> ```
> markers = [
>     "slow_main: marks slower tests run at merge to main",
>     "long_local: marks long tests run locally",
>     "realization_counting: marks tests using extra realization-counting",
>     "meshing: marks tests using extra meshing"
> ]
> ```
> 
> ?

I like it!

matthiashimmelmann (approved):

PR #159, opened 2024-11-20
Feature tikz

ggmath:
Basic functionality on tikz strings for graphs and frameworks added


pyrigi/graph.py
```diff
@@ -2181,6 +2181,225 @@ def random_framework(self, dim: int = 2, rand_range: Union(int, List[int]) = Non
...
         return Framework.Random(self, dim, rand_range)
 
+    @doc_category("Other")
+    def to_tikz(
+        self,
+        layout_type: str = "spring",
+        placement: dict[Vertex, Point] = None,
+        vertex_style: Union(str, dict[str : list[Vertex]]) = "gvertex",
+        edge_style: Union(str, dict[str : list[Edge]]) = "edge",
+        label_style: str = "labelsty",
```
Legersky, line 2191: `label_style` is not used in the method

pyrigi/framework.py
```diff
@@ -436,6 +436,112 @@ def plot(
 
         self.plot2D(**kwargs)
 
+    @doc_category("Other")
+    def to_tikz(
```
Legersky, line 440: The method should work only for 2-dimensional frameworks.
Currently, calling it on a 1-dim framework raises `IndexError` while for 3-dim it silently takes the first two coordinates.
I would just raise `ValueError` if the dimension is different from 2.

pyrigi/framework.py
```diff
@@ -436,6 +436,112 @@ def plot(
...
+        vertex_style: Union(str, dict[str : list[Vertex]]) = "fvertex",
+        edge_style: Union(str, dict[str : list[Edge]]) = "edge",
+        label_style: str = "labelsty",
+        figure_opts: str = "",
+        vertex_in_labels: bool = False,
+        vertex_out_labels: bool = False,
+        default_styles: bool = True,
+    ) -> str:
+        r"""
+        Create a tikz code for the framework.
```
MatteoGallet, line 451: I would propose to change `Create a tikz code for the framework.` -> `Create TikZ code for the framework.`

pyrigi/graph.py
```diff
@@ -2181,6 +2181,223 @@ def random_framework(self, dim: int = 2, rand_range: Union(int, List[int]) = Non
...
+        figure_opts: str = "",
+        vertex_in_labels: bool = False,
+        vertex_out_labels: bool = False,
+        default_styles: bool = True,
+    ) -> str:
+        r"""
+        Create a tikz code for the graph.
```
MatteoGallet, line 2198: I would propose to change `Create a tikz code for the graph.` -> `Create TikZ code for the graph.`

Legersky (approved):

PR #166, opened 2024-12-04
Refactor minimal rigid subgraphs

matthiashimmelmann:
+ Removed `min_rigid_subgraphs` due to lacking usefulness
+ Added tests
+ Added definition

### DIFF
--- a/doc/math/rigidity/generic.md
+++ b/doc/math/rigidity/generic.md
@@ -42,8 +42,8 @@ A graph $G = (V, E)$ is minimally (generically) $2$-rigid if and only if $G$ is 
 :label: def-rigid-components
 
 Let $G$ be a graph, let $d \in \NN$.
-The _rigid components_ of $G$ are the maximal vertex-induced subgraphs 
-of $G$ with at least $2$ vertices that are {prf:ref}`rigid <def-gen-rigid>`.
+The _d-rigid components_ of $G$ are the maximal vertex-induced subgraphs 
+of $G$ that are {prf:ref}`d-rigid <def-gen-rigid>`.
 
 {{pyrigi_crossref}} {meth}`~.Graph.rigid_components`
 :::
--- a/doc/math/rigidity/generic.md
+++ b/doc/math/rigidity/generic.md
@@ -43,7 +43,7 @@ A graph $G = (V, E)$ is minimally (generically) $2$-rigid if and only if $G$ is 
 
 Let $G$ be a graph, let $d \in \NN$.
 The _rigid components_ of $G$ are the maximal vertex-induced subgraphs 
-of $G$ with at least $d+1$ vertices that are rigid.
+of $G$ with at least $2$ vertices that are rigid.
 
 {{pyrigi_crossref}} {meth}`~.Graph.rigid_components`
 :::
--- a/doc/math/rigidity/generic.md
+++ b/doc/math/rigidity/generic.md
@@ -38,3 +38,12 @@ A graph $G = (V, E)$ is minimally (generically) $2$-rigid if and only if $G$ is 
 :::
 
 
+:::{prf:definition} Rigid components in $\RR^d$
+:label: def-rigid-components
+
+Let $G$ be a graph, let $d \in \NN$.
+The _rigid components_ of $G$ are the maximal vertex-induced subgraphs 
+of $G$ with at least $d+1$ vertices that are rigid.
+
+{{pyrigi_crossref}} {meth}`~.Graph.rigid_components`
+:::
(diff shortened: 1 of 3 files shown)


doc/math/rigidity/generic.md
```diff
@@ -38,3 +38,12 @@ A graph $G = (V, E)$ is minimally (generically) $2$-rigid if and only if $G$ is
 :::
 
 
+:::{prf:definition} Rigid components in $\RR^d$
+:label: def-rigid-components
+
+Let $G$ be a graph, let $d \in \NN$.
+The _rigid components_ of $G$ are the maximal vertex-induced subgraphs 
+of $G$ with at least $d+1$ vertices that are rigid.
```
Legersky, line 46: I think the standard is no restriction on the number of vertices, i.e., every edge is in some rigid component.

doc/math/rigidity/generic.md
```diff
@@ -38,3 +38,12 @@ A graph $G = (V, E)$ is minimally (generically) $2$-rigid if and only if $G$ is
 :::
 
 
+:::{prf:definition} Rigid components in $\RR^d$
+:label: def-rigid-components
+
+Let $G$ be a graph, let $d \in \NN$.
+The _rigid components_ of $G$ are the maximal vertex-induced subgraphs 
+of $G$ with at least $2$ vertices that are rigid.
```
MatteoGallet, line 46: Could you please change

```
that are rigid
```

by

```
that are {prf:ref}`rigid <def-gen-rigid>`
```

pyrigi/graph.py
```diff
@@ -1880,35 +1880,31 @@ def is_Rd_closed(self, dim: int = 2) -> bool:
...
-        missing definition, tests
+        :prf:ref:`Rigid components <def-rigid-components>`
 
         Notes
         -----
         We only return nontrivial subgraphs, meaning that there need to be at
-        least ``dim+1`` vertices present. If the graph itself is rigid, it is clearly
-        maximal and is returned.
+        least ``2`` vertices present. If the graph itself is rigid, it is
+        clearly maximal and is returned.
```
Legersky, line 1895: Is there a reason to return only non-trivial ones? If a graph has an isolated vertex, then this vertex is a rigid component so I think it should be returned.

doc/math/rigidity/generic.md
```diff
@@ -38,3 +38,12 @@ A graph $G = (V, E)$ is minimally (generically) $2$-rigid if and only if $G$ is
 :::
 
 
+:::{prf:definition} Rigid components in $\RR^d$
+:label: def-rigid-components
+
+Let $G$ be a graph, let $d \in \NN$.
+The _rigid components_ of $G$ are the maximal vertex-induced subgraphs 
+of $G$ with at least $2$ vertices that are {prf:ref}`rigid <def-gen-rigid>`.
```
Legersky, line 46: There should be `$d$-rigid`. Moreover, I would drop the requirement to have at least two vertices, see the other comment.

pyrigi/graph.py
```diff
@@ -1920,19 +1916,17 @@ def max_rigid_subgraphs(self, dim: int = 2) -> List[Graph]:
...
+            return [list(self)]
         rigid_subgraphs = {
             tuple(vertex_subset): True
-            for r in range(dim + 1, self.number_of_nodes() - 1)
+            for r in range(2, self.number_of_nodes()-1)
             for vertex_subset in combinations(self.nodes, r)
-            if self.subgraph(vertex_subset).is_rigid(dim)
-        }
+            if self.subgraph(vertex_subset).is_rigid(dim, combinatorial=(dim < 3))
+        } | {tuple([v]): True for v in self.nodes}
```
Legersky, line 1929: We could just put instead
```py
if self.number_of_nodes() == 1:
    return [self.vertex_list()]
```
above, since we here it is guaranteed that the graph is connected that to the condition above.

pyrigi/graph.py
```diff
@@ -1913,20 +1913,22 @@ def rigid_components(self, dim: int = 2) -> List[List[Vertex]]:
...
+        if nx.is_empty(self):
+            return([])
         if not nx.is_connected(self):
             res = []
             for comp in nx.connected_components(self):
                 res += self.subgraph(comp).rigid_components(dim)
             return res
 
-        if self.is_rigid(dim, combinatorial=(dim < 3)):
+        if self.is_rigid(dim, combinatorial=(dim < 3)) or self.number_of_nodes() == 1:
```
Legersky, line 1924: Actually, here the first condition should suffice, but we have a bug in `is_rigid` !
```
>>> G = Graph.from_vertices([0])
>>> G.is_rigid()
False
```

matthiashimmelmann: Yeah, I was confused about that as well. It is this part:

```
deficiency = -(2 * self.number_of_nodes() - 3) + self.number_of_edges()
if deficiency < 0:
    return False
```

Does the deficiency argument hold for arbitrary graphs or do we need to give special treatment to the complete graphs until K_{d-1}?

Legersky: it needs special treatment for single vertex graphs, as you did. I just moved the check directly to the 2d case, since in other dimensions the treatment is different/necessary for more complete graphs.

Legersky (approved):

PR #170, opened 2024-12-10
Refactor random rigidity

ggmath:
Allow user to chose probability on rigidity test with random matrix and redundancy test, where in the later, the probability is just passed on to the rigidity test.


pyrigi/graph.py
```diff
@@ -1540,25 +1611,33 @@ def is_rigid(self, dim: int = 2, combinatorial: bool = True) -> bool:
...
                 return False
             else:
                 self._build_pebble_digraph(2, 3)
                 return (
                     self._pebble_digraph.number_of_edges()
-                    == 2 * self.number_of_nodes() - 3
+                    == 2 * n - 3
                 )
         elif not combinatorial:
+            N = int((n * dim - math.comb(dim + 1, 2)) / prob)
```
Legersky, line 1635: This computation of `N` does not work for graphs with too little vertices.

Legersky (approved):

PR #171, opened 2024-12-11
ParametricMotion animation using SVG

rubesjan:


pyrigi/motion.py
```diff
@@ -106,10 +118,155 @@ def __str__(self) -> str:
...
+                tan_placement = []
+                for coord in placement:
+                    tan_placement.append(
+                        coord.replace(self._parameter, sp.sympify("tan(t)"))
+                    )
+                mot[v] = tan_placement
+            mot = ParametricMotion(deepcopy(self._graph), mot, (-np.pi / 2, np.pi / 2))
+            return mot.animate(
+                width, height, filename, sampling, show_labels, vertex_size, length
+            )
```
Legersky, line 219: Is it fast enough? Doing it just numerically during sampling would avoid all the symbolic checks when creating new ParametricMotion.

pyrigi/motion.py
```diff
@@ -106,10 +118,155 @@ def __str__(self) -> str:
...
+            animation. A higher value results in a smoother and more accurate
+            representation of the motion, while a lower value can speed up rendering
+            but may lead to a less precise or jerky animation. This parameter controls
+            the resolution of the animation's movement by setting the density of
+            sampled data points between keyframes or time steps.
+        show_labels:
+            If true, vertices will have a number label.
+        vertex_size:
+            Size of vertices in the animation.
+        length:
```
Legersky, line 199: Could you change this to `duration`? (Since this is rigidity theory package, 'lenght' reminds distance).

pyrigi/motion.py
```diff
@@ -106,10 +118,155 @@ def __str__(self) -> str:
...
+            animation = f'<animate href="#edge{v_to_int[u]}-{v_to_int[v]}" '
+            animation += f'attributeName="d" dur="{length}s" '
+            animation += 'repeatCount="indefinite" calcMode="linear" '
+            animation += f'values="{positions_str}"/>'
+            svg = svg + "\n" + animation
+        svg = svg + "\n</svg>"
+
+        if filename is not None:
+            with open(filename, "wt") as file:
+                file.write(svg)
```
Legersky, line 271: Could you add `.svg` in case it is not part of the name?

pyrigi/motion.py
rubesjan: It should be fast but I agree that it is unnecessary. I changed the code so no new ParametricMotion is created.

pyrigi/motion.py
rubesjan: Done

pyrigi/motion.py
rubesjan: That's a good idea. Code is updated

Legersky (approved):

PR #174, opened 2024-12-12
Added the methods `is_trivial_flex` and `is_nontrivial_flex`

matthiashimmelmann:
+ `is_trivial_flex` is an alias for `is_dict_trivial_inf_flex` and `is_vector_trivial_inf_flex`
+ `is_nontrivial_flex` is an alias for `is_dict_nontrivial_inf_flex` and `is_vector_nontrivial_inf_flex`
+ We also add the appropriate tests
+ Unified parameter lists


pyrigi/data_type.py
```diff
@@ -35,9 +36,9 @@
 A Point is a Sequence of Coordinates whose length is the dimension of its affine space.
 """
 
-Stress = Sequence[int | float]
+Stress = Sequence[int | float | str]
```
matthiashimmelmann, line 39: A stress can be symbolic, too.

pyrigi/data_type.py
```diff
@@ -50,20 +51,29 @@ def point_to_vector(point: Point) -> Matrix:
...
+            )
+        return point if (point.shape[1] == 1) else point.transpose()
 
     if not isinstance(point, Sequence) or isinstance(point, str):
         raise TypeError("The point must be a Sequence of Coordinates.")
 
     try:
         res = Matrix(point)
-    except SympifyError as e:
+    except Exception as e:
```
matthiashimmelmann, line 74: We should catch all possible exceptions that come about when converting to `Matrix`. Especially, because we cannot possibly catch all exceptions in the lines before.

pyrigi/data_type.py
```diff
@@ -50,20 +51,29 @@ def point_to_vector(point: Point) -> Matrix:
     """
     Return point as single column sympy Matrix.
     """
-
-    if isinstance(point, MatrixBase):
-        if point.shape[1] != 1:
+    if isinstance(point, MatrixBase) or isinstance(point, np.ndarray):
+        if (
```
matthiashimmelmann, line 55: We should be more lenient with the input shape.

Legersky (approved):

PR #186, opened 2024-12-27
Added a plotting method for equilibrium stresses

matthiashimmelmann:
+ Added a plotting method for equilibrium stresses.
+ Added improved 1D (and collinear) plotting capabilities via the `connection_style` Parameter.

### DIFF
--- a/doc/userguide/tutorials/plotting.md
+++ b/doc/userguide/tutorials/plotting.md
@@ -187,8 +187,15 @@ F.plot(
 
 ### Collinear Configurations
 
-For collinear configurations and frameworks in $\RR$, we can visualize the edges as arcs
-in $\RR^2$ using the keyword ``connection_style``. This parameter can be specified in
+For collinear configurations and frameworks in $\RR$, the edges are automatically visualized 
+as arcs in $\RR^2$ 
+
+```{code-cell} ipython3
+F = Framework.Complete([[0],[1],[2]])
+F.plot()
+```
+
+Using the keyword ``connection_style``, we are able to specify the pitch of the individual arcs. This parameter can be specified in
 radians as a ``float`` if the same pitch for every arc is desired and a ``list[float]`` or a
 ``dict[Edge, float]`` if the pitch is supposed to be provided for each arc individually.
 
--- a/doc/userguide/tutorials/plotting.md
+++ b/doc/userguide/tutorials/plotting.md
@@ -220,7 +220,8 @@ F.plot(curved_edges=True, connection_style={(i,i+1): 0.15 for i in range(7)} | {
 It is possible to include infinitesimal flexes in the plot. With the keyword
 `inf_flex=n`, we can pick the `n`-th nontrivial infinitesimal flex from
 a basis of the rigidity matrix's kernel. There are several keywords that allow
-us to alter the style of the drawn arrows.
+us to alter the style of the drawn arrows. A full list of the optional plotting
+parameters can be found in the API reference: {meth}`~.Graph.plot`.
 
 ```{code-cell} ipython3
 G = Graph([[0, 1], [0, 2], [1, 2], [2, 3], [2, 4], [3, 4]])
@@ -264,7 +265,12 @@ Analogous to the way that infinitesimal flexes can be visualized (see the previo
 section), a `stress` can be provided either as the `n`-th equilibrium stress, as a
 specific `stress` given by a `Matrix` or alternatively as a `Dict[Edge, Coordinate]`. 
 It is internally checked, whether the provided stress lies in the cokernel of the
-rigidity matrix. 
+rigidity matrix. We can specify the positions of the stress labels using the keyword
+`stress_label_pos`, which can either be set for each edge as the same `float` from $[0,1]$
+or individually using a `dict[DirectedEdge, float]`. This `float` specifies the position on
+the line segment given by the edge. The missing edges labels are automatically
+centered on the edge. A full list of the optional plotting parameters can be found in
+the API reference: {meth}`~.Graph.plot`. 
 
 ```{code-cell} ipython3
 F = frameworks.Frustum(3)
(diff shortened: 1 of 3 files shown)


pyrigi/framework.py
```diff
@@ -1232,7 +1288,47 @@ def pinned_rigidity_matrix(
...
+        """
+        stress_edge_list = [tuple(e) for e in list(dict_stress.keys())]
+        self._check_edge_order(stress_edge_list)
+        graph_edge_list = [tuple(e) for e in self._graph.edge_list()]
+        dict_to_list = []
+
+        for e in graph_edge_list:
+            if e not in stress_edge_list:
+                raise ValueError(f"Edge {e} must be in the dictionary `dict_stress`.")
+            dict_to_list += [dict_stress[e]]
```
Legersky, line 1324: I think this fails if an edge uv is provided as `(u,v)` in the `dict_stress` but as `(v,u)` from `edge_list()`.

pyrigi/graph.py
```diff
@@ -2765,21 +2795,105 @@ def plot(
...
+                    tuple(e): connection_style for e in self.edge_list()
+                }
+            elif isinstance(connection_style, list):
+                if not self.number_of_edges() == len(connection_style):
+                    raise AttributeError(
+                        "The provided `connection_style` doesn't have the correct length."
+                    )
+                connection_style = {
+                    tuple(self.edge_list()[i]): connection_style[i]
+                    for i in range(self.number_of_edges())
```
Legersky, line 2828: Can this be simplified as
```python
tuple(e):style for e, style in zip(self.edge_list(), connection_style)
```
?

matthiashimmelmann: ✅

pyrigi/framework.py
matthiashimmelmann: This was unnecessary to begin with, since the edge order is checked a few lines before, so both lists contain the same edges.

doc/userguide/tutorials/plotting.md
```diff
@@ -185,7 +185,30 @@ F.plot(
...
-## Infinitesimal Flexes
+### Collinear Configurations
+
+For collinear configurations and frameworks in $\RR$, we can visualize the edges as arcs
+in $\RR^2$ using the keyword ``connection_style``. This parameter can be specified in
+radians as a ``float`` if the same pitch for every arc is desired and a ``list[float]`` or a
+``dict[Edge, float]`` if the pitch is supposed to be provided for each arc individually.
+
+```{code-cell} ipython3
+F = Framework.Complete([[1],[3],[0],[2]])
```
Legersky, line 196: Shall we first show that for 1-dimensional framework, the edges are curved automatically? i.e. `F.plot()`?

pyrigi/graph.py
```diff
@@ -2678,10 +2678,11 @@ def layout(self, layout_type: str = "spring") -> dict[Vertex, Point]:
             raise ValueError(f"layout_type {layout_type} is not supported.")
 
     @doc_category("Other")
-    def plot(
+    def plot(  # noqa: C901
         self,
         placement: dict[Vertex, Point] = None,
         inf_flex: dict[Vertex, Sequence[Coordinate]] = None,
+        stress: dict[Edge, Coordinate] = None,
```
Legersky, line 2685: Are also other formats possible?

pyrigi/graph.py
```diff
@@ -2857,6 +2980,23 @@ def plot(
                 **kwargs,
             )
 
+        if stress is not None:
+            nonzero_stress = next(
+                (w for _, w in stress.items() if not sympify(w).evalf(10) == 0), None
+            )
+            if nonzero_stress is None:
+                raise ValueError("The provided stress only contains zeros.")
+            _stress = {k: w / nonzero_stress for k, w in stress.items()}
```
Legersky, line 2989: I don't see a reason to normalize the stress by the first nonzero entry, especially if it was explicitly provided.

matthiashimmelmann: The problem I am trying to fix occurs with complicated or random symbolic coordinates. See for example

```
G = graphDB.Complete(4)
F = Framework.Random(G)
F.plot(stress=0)
```

I am not happy with the current result though, so I am open for suggestions.

pyrigi/graph.py
matthiashimmelmann: No, I expect that the `stress` is already transformed to a `dict` at this point. I will fix this, once we move all plotting the `Framework`.

pyrigi/graph.py
```diff
@@ -2756,6 +2764,16 @@ def plot(
...
         flex_arrowsize:
             Size of the arrowhead's length and width.
+        stress_color:
+            Color of the font used to label the edges with stresses.
+        stress_fontsize:
+            Fontsize of the stress labels.
+        stress_label_pos:
+            Position of the stress label along the edge. `float` numbers
+            from the interval `[0,1]` are allowed. `0` represents the head
+            of the edge, `0.5` the center and `1` the edge's tail.
```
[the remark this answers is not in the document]
matthiashimmelmann: I guess they are defined by the internal vertex order of `networkx`. I tried coming up with a reasonable way to work around it, but couldn't really find one ad hoc.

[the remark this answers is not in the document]
Legersky: I guess the same. Do they need to be positioned differently than in the middle in some situation?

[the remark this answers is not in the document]
matthiashimmelmann: Yes, consider for example the standard way of visualizing `frameworkDB.Complete(4)`. Then, the central stresses overlap.

doc/userguide/tutorials/plotting.md
matthiashimmelmann: ✅

pyrigi/graph.py
Legersky: I see: I think by default the displayed stress should be kept as it is (e.g. symbolic when it was symbolic). There could be an option to normalize it/make numeric.

matthiashimmelmann: That's a good idea. I'll adress it in the next commit.

matthiashimmelmann: ✅

matthiashimmelmann:
> Shall we extend `edge_list` with parameter `as_tuples` (default `False`) to avoid all the manual retyping?

I don't understand this comment. can you elaborate?

PR #187, opened 2024-12-30
Feature plot 3d

Maaffaa:
Plot 3D for frameworks.


pyrigi/framework.py
```diff
@@ -477,6 +480,286 @@ def plot2D(  # noqa: C901
...
+        ).items():
+            placement[vertex] = np.dot(projection_matrix, np.array(position))
+
+        self._plot_with_3D_realization(placement, **kwargs)
+
+    @doc_category("Other")
+    def _plot_with_3D_realization(
+        self,
+        realization: dict[Vertex, Point],
+        inf_flex: dict[Vertex, Sequence[Coordinate]] = None,
```
matthiashimmelmann, line 719: `realization` and `inf_flex` are not used in `_plot_with_3D_realization`.

pyrigi/framework.py
```diff
@@ -477,6 +480,286 @@ def plot2D(  # noqa: C901
...
+        # Create a figure for the rapresentation of the framework
+        fig = plt.figure(figsize=(10, 10))
+        ax = fig.add_subplot(111, projection="3d")
+
+        pos = self.realization(as_points=True, numerical=True)
+        # Draw the vertices as points in the 3D enviroment
+        x_nodes = [pos[node][0] for node in self._graph.nodes]
+        y_nodes = [pos[node][1] for node in self._graph.nodes]
+        z_nodes = [pos[node][2] for node in self._graph.nodes]
+        ax.scatter(x_nodes, y_nodes, z_nodes, c=vertex_color, s=200)
```
matthiashimmelmann, line 746: Some vertices are displayed transparently. Is this a `matplotlib` feature? Should we play a bit with this to highlight the position on the `z`-axis?

pyrigi/framework.py
```diff
@@ -477,6 +480,286 @@ def plot2D(  # noqa: C901
...
+        placement = {}
+        for vertex, position in self.realization(
+            as_points=False, numerical=True
+        ).items():
+            placement[vertex] = np.dot(projection_matrix, np.array(position))
+
+        self._plot_with_3D_realization(placement, **kwargs)
+
+    @doc_category("Other")
+    def _plot_with_3D_realization(
```
matthiashimmelmann, line 716: Would you mind writing an interactive 3D-plotting tutorial for Jupyter Notebooks using `%matplotlib widget`?

pyrigi/framework.py
```diff
@@ -477,6 +480,286 @@ def plot2D(  # noqa: C901
...
+            return self.plot2D(**kwargs)
+
+        if self._dim == 3 and not animation:
+            placement = self.realization(as_points=True, numerical=True)
+            self._plot_with_3D_realization(
+                placement, inf_flex=inf_flex_pointwise, **kwargs
+            )
+            return
+
+        elif self._dim == 3 and animation:
```
matthiashimmelmann, line 571: Please put the rotation sequence in a new method called `_animate_rotation_around_z_axis` or similar. Allow the user to specify the rotation sequence.

pyrigi/framework.py
```diff
@@ -477,6 +480,286 @@ def plot2D(  # noqa: C901
...
+        self._plot_with_3D_realization(placement, **kwargs)
+
+    @doc_category("Other")
+    def _plot_with_3D_realization(
+        self,
+        realization: dict[Vertex, Point],
+        inf_flex: dict[Vertex, Sequence[Coordinate]] = None,
+        vertex_color="#ff8c00",
+        edge_width=1.5,
+        **kwargs,
```
matthiashimmelmann, line 722: Allow the user to specify `vertex_size: int = 300`, `vertex_shape: str = "o"` (if possible), `edge_color: str | List[List[Edge]] | Dict[str : List[Edge]] = "black"`, `edge_style: str = "solid"` and `font_color: str = "whitesmoke"`.

pyrigi/misc.py
```diff
@@ -88,6 +88,44 @@ def generate_two_orthonormal_vectors(dim: int, random_seed: int = None) -> Matri
...
+        The dimension in which the vectors are generated.
+    random_seed:
+        Seed for generating random vectors.
+        When the same value is provided, the same vectors are generated.
+    """
+
+    if random_seed is not None:
+        np.random.seed(random_seed)
+
+    matrix = np.random.randn(dim, 3)
```
matthiashimmelmann, line 109: I wonder if it would be easier to generate the QR-decomposition of `matrix` and then take `Q` as your orthonormal basis. Since this method has a robust implementation via Householder transforms, you could get rid of the shenanigans with the resampling (and the theoretically infinite loop).

pyrigi/framework.py
Maaffaa: I'm sorry, what do you mean by "rotation sequence"?

matthiashimmelmann: Currently, the rotation is performed around the `z`-axis:

```
rotation_matrix = np.array(
    [
        [np.cos(angle), -np.sin(angle), 0],
        [np.sin(angle), np.cos(angle), 0],
        [0, 0, 1],
    ]
)
```

Maybe you can allow the user to input a parametric rotation matrix as a function. E.g. 

```
def rotation_matrix(angle) -> Matrix:
  return np.array(
      [
          [np.cos(angle), -np.sin(angle), 0],
          [np.sin(angle), np.cos(angle), 0],
          [0, 0, 1],
      ]
  )
```

and then distinguish `isinstance(rotation_matrix, types.FunctionType)` from `isinstance(rotationmatrix, Matrix | np.ndarray)`?

Maaffaa: So you mean something of this type?
     
        # Function to update data at each frame
        def update(frame, rotation_matrix, **kargs):
            if rotation_matrix is None:
                angle = frame * np.pi / 50
                # Rotation of vertices around the z-axis
                rotation_matrix = np.array(
                    [
                        [np.cos(angle), -np.sin(angle), 0],
                        [np.sin(angle), np.cos(angle), 0],
                        [0, 0, 1],
                    ]
                )
            elif isinstance(rotation_matrix, types.FunctionType):
                rotation_matrix = rotation_matrix(frame * np.pi / 50)
            
            rotated_vertices = vertices.dot(rotation_matrix.T)

            # Update vertices positions

matthiashimmelmann: Yes, that's exactly what I had in mind. And then you can raise a `TypeError` if `roation_matrix` is neither `None` nor `types.FunctionType`.

Maaffaa: I just noticed that this way of acting has no sense since the `rotation_matrix` value changes at each frame, it actually depends on it. So the only input the user could write as `rotation_matrix` is a function, I think. I'm going to change the code in this direction.

matthiashimmelmann: Hey @Maaffaa , I just realized there is a method `rotate2D` in `Framework`. Would it be possible to add the method 
```rotate3D(self, angle: float, axis: Vector, inplace: bool = True) -> None | Framework``` 
in the format we discussed above? It would theoretically be possible be use that method in your `update` function. If that is too much work, just let me know and I'll write that function myself. 

By the way, thanks a lot for your great work!

File: pyrigi/framework.py (hunk 1 above)
Comment (Maaffaa) [reply pyrigi/framework.py:719]: You are right, I've just made this correction.

File: pyrigi/framework.py (hunk 5 above)
Comment (Maaffaa) [reply pyrigi/framework.py:722]: I have just done all these modifications except the `edge_color`: I am trying to see if there is an easy way to make the code accepting a list of different colours.

File: pyrigi/misc.py (hunk 6 above)
Comment (Maaffaa) [reply pyrigi/misc.py:109]: That's a good idea. I implemented it and checked that it takes the same amount of time. I used the QR decomposition of numpy.

File: pyrigi/framework.py (hunk 4 above)
Comment (matthiashimmelmann) [reply pyrigi/framework.py:652]: I realized that it would probably be good to also allow the user to specify a rotation axis as a vector. Then, there would be three options: `None | Vector | Function`. Probably vector can also replace `Function`, since all rotations in $\RR$ are uniquely determined by their axis of rotation on $S^2$ and the planar rotation around that axis. Sorry, I only made up my mind about this now...

File: pyrigi/framework.py [hunk 7]
```diff
@@ -479,31 +483,440 @@ def plot2D(  # noqa: C901
...
+            If a dictionary from colors to a list of edge is given,
+            edges are colored accordingly.
+            The edges missing in the partition/dictionary, are colored black.
+        edge_style:
+            Edge line style: ``-``/``solid``, ``--``/``dashed``,
+            ``-.``/``dashdot`` or ``:``/``dotted``. By default '-'.
+        equal_aspect_ratio:
+            Determines whether the aspect ratio of the plot is equal in all space
+            directions or whether it is adjusted depending on the framework's size
+            in `x`, `y` and `z`-direction individually.
```
Legersky, line 824: This is not visible in the documentation as the method is non-public.

Legersky: (can be solved when refactoring the main plotting functionality from `Graph` to `Framework`)

Legersky (approved):

PR #188, opened 2024-12-31
Refactoring `is_Rd_circuit`

matthiashimmelmann:
+ Added implementation for `dim=3`
+ Added Tests
+ Minor refactoring

### DIFF
--- a/pyrigi/graph.py
+++ b/pyrigi/graph.py
@@ -6,7 +6,7 @@ from __future__ import annotations
 
 from copy import deepcopy
 from itertools import combinations
-from typing import List, Dict, Union, Iterable
+from typing import List, Dict, Iterable
 
 import networkx as nx
 import matplotlib.pyplot as plt
@@ -927,7 +927,7 @@ class Graph(nx.Graph):
     @doc_category("Generic rigidity")
     def extension_sequence(
         self, dim: int = 2, return_solution: bool = False
-    ) -> Union[List[Graph], bool]:
+    ) -> List[Graph] | bool:
         """
         Check the existence of a sequence of
         :prf:ref:`0 and 1-extensions <def-k-extension>`.
@@ -1866,7 +1866,7 @@ class Graph(nx.Graph):
                     "There must be at least one stress but none was found."
                 )
 
-    @doc_category("Partially implemented")
+    @doc_category("Rigidity Matroid")
     def is_Rd_dependent(
         self, dim: int = 2, use_precomputed_pebble_digraph: bool = False
     ) -> bool:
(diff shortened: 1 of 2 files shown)


pyrigi/graph.py
```diff
@@ -1928,9 +1931,10 @@ def is_Rd_independent(
                 2, 3, use_precomputed_pebble_digraph=use_precomputed_pebble_digraph
             )
 
-        raise NotImplementedError()
+        F = self.random_framework(dim=dim)
```
Legersky, line 1934: This silently switches to a randomized algorithm, we should keep it in mind when dealing with randomized algorithms in general, please, write it as a todo.

pyrigi/graph.py
```diff
@@ -1982,22 +1981,39 @@ def is_Rd_circuit(
...
+                tuple(edge) for edge in max_sparse_subgraph.edges()
+            ]
+            remaining_edges = [
+                edge
+                for edge in self.edges()
+                if not (
+                    tuple(edge) in max_sparse_subgraph_edges
+                    or tuple([edge[1], edge[0]]) in max_sparse_subgraph_edges
+                )
+            ]
```
Legersky, line 1994: Can it be simplified to
```python
remaining_edges = [ edge for edge in self.edges() if not max_sparse_subgraph.has_edge(*e)]
```
?

pyrigi/graph.py
```diff
@@ -1982,22 +1981,39 @@ def is_Rd_circuit(
...
                 )
                 == self.number_of_nodes()
             )
 
-        raise NotImplementedError()
+        if not self.is_Rd_dependent(dim=dim):
+            return False
+        for edge in self.edge_list():
+            G = deepcopy(self)
+            G.remove_edge(*edge)
```
Legersky, line 2013: To avoid creating many copies, which is usually slow, a single copy can be constructed before the loop and then `edge` is removed and added back after the independence check.

Legersky:
Please, could you improve also the docstrings of the touched methods?

pyrigi/graph.py
matthiashimmelmann: ✅

pyrigi/graph.py
```diff
@@ -1982,22 +1981,39 @@ def is_Rd_circuit(
...
+                        v=remaining_edges[0][1],
                     )
                 )
                 == self.number_of_nodes()
             )
 
-        raise NotImplementedError()
+        if not self.is_Rd_dependent(dim=dim):
+            return False
+        for edge in self.edge_list():
```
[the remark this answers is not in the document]
matthiashimmelmann: ✅

pyrigi/graph.py
matthiashimmelmann: ✅

Legersky (approved):

PR #190, opened 2025-01-01
Refactor the codebase: consistent using of dim, type hinting

matthiashimmelmann:
+ Added missing 1-line comments
+ `d` -> `dim` for consistency
+ PEP 545: `Dict`, `List`, `Union` are deprecated. Now `dict`, `list`, `|` are generic types. For the input type, we should be able to deal with a `Sequence` instead of specifying `list | tuple`, while for the output type, we should specify exactly what the output is, e.g. `list`.

### DIFF
--- a/pyrigi/graph.py
+++ b/pyrigi/graph.py
@@ -665,23 +665,23 @@ class Graph(nx.Graph):
         self,
         vertices: Sequence[Vertex],
         new_vertex: Vertex = None,
-        dim: int = 2,
+        d: int = 2,
         inplace: bool = False,
     ) -> Graph:
         """
-        Return a :prf:ref:`dim-dimensional 0-extension <def-k-extension>`.
+        Return a :prf:ref:`d-dimensional 0-extension <def-k-extension>`.
 
         Parameters
         ----------
         vertices:
             A new vertex will be connected to these vertices.
             All the vertices must be contained in the graph
-            and there must be ``dim`` of them.
+            and there must be ``d`` of them.
         new_vertex:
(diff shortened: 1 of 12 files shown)


MatteoGallet (commented):
I do like the use of `dim` over `d`. On the other hand, as it is in the proposed branch, we still use expressions as "d-dimensional" or "d-rigid" or "the parameter d" in the docstrings; this makes it a bit non-homogenous. What do you think @Legersky  and @ggmath ?

matthiashimmelmann:
> I do like the use of `dim` over `d`. On the other hand, as it is in the proposed branch, we still use expressions as "d-dimensional" or "d-rigid" or "the parameter d" in the docstrings; this makes it a bit non-homogenous. What do you think @Legersky and @ggmath ?

I have also noticed that, but I guess it is technically ok to have different standards for code and documentation. They have different goals: OOP code should be oriented to natural language. Let's discuss this in more detail though, as this issue occurs not only for dimension!

pyrigi/graph.py
```diff
@@ -1871,6 +1872,8 @@ def is_Rd_dependent(
         self, dim: int = 2, use_precomputed_pebble_digraph: bool = False
     ) -> bool:
         """
+        Checks whether the graph's edge set is dependent in the d-rigidity matroid.
```
Legersky, line 1875: Could you add the link to the corresponding definition? Please, use imperative and `dim` instead of `d`.

Legersky: (We can do this in a separate PR as well.)

PR #191, opened 2025-01-03
`inf_flex` 3d Plotting

matthiashimmelmann:
Allows the plotting of infinitesimal flexes in 3D figures. Depends on #187 .

After this PR, we should unify the plotting and add it to `Framework` rather than doing it all in `Graph`.


pyrigi/framework.py
```diff
@@ -929,19 +961,160 @@ def _plot_with_3D_realization(
...
+                    va="center",
+                    alpha=1,
+                )
+        self.plot_inf_flex(ax, inf_flex, **kwargs)
+
         plt.tight_layout()
         plt.show()
 
+    @doc_category("Other")
+    def plot_inf_flex(  # noqa: C901
```
Legersky, line 982: Does it make sense to have this as a public method?

pyrigi/framework.py
```diff
@@ -729,7 +735,7 @@ def update(frame):
     @doc_category("Other")
     def plot3D(
         self,
-        coordinates: Sequence[int] = None,
+        coordinates: Sequence[Coordinate] = None,
```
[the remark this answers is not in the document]
matthiashimmelmann: Sorry, you're right.
✅

pyrigi/framework.py
Legersky: In that case I would prefer to make it non-public since changing it later to public if needed is much better than the other way around. Documenting the parameters is something we will have to unify anyway.

Legersky (approved):

PR #207, opened 2025-01-24
Migrate `animate3D` to the `Motion` class

matthiashimmelmann:
+ Unified the plotting styles of `animate2D` and `animate3D` into `matplotlib.FuncAnimation`. 
+ Allowed the keyword `plot_style`


pyrigi/framework.py
```diff
@@ -468,67 +467,58 @@ def plot2D(
             )
 
     @doc_category("Plotting")
-    def animate3D(
+    def animate_rotating_framework3D(
```
MatteoGallet, line 470: Could we maybe avoid `framework` in the name of the method, since anyway it is a method of the `Framework` class? Maybe something like `animate3D_rotating`?

pyrigi/motion.py
```diff
@@ -40,149 +50,214 @@ def graph(self) -> Graph:
...
+        plot_style: PlotStyle,
+        edge_coloring: Sequence[Sequence[Edge]] | dict[str, Sequence[Edge]] = None,
+        delay: int = 75,
+        **kwargs,
+    ) -> Any:
         """
-        Fix the edge ``fixed_edge`` for every entry of ``realizations``.
+        Animate the continuous motion.
+
+        See :class:`~.PlotStyle2D` for a list of possible visualization keywords.
```
Legersky, line 115: -> See ``:class:`~.PlotStyle` `` or ``:class:`~.PlotStyle3D` ``

pyrigi/motion.py
```diff
@@ -346,15 +460,14 @@ def __init__(
         Creates an instance.
         """
 
-        super().__init__(graph)
+        super().__init__(graph, len(motion[len(list(motion.values())[0])]))
```
Legersky, line 463: Why not just `len(list(motion.values()))`?

pyrigi/motion.py
```diff
@@ -515,6 +628,14 @@ class ApproximateMotion(Motion):
         distance is at least ``turning_threshold`` times as large as the distance
         of the negative infinitesimal flex, then the latter one is chosen instead.
         If instead the animation is too slow, consider increasing this value.
+    fixed_edge:
+        The edge of the underlying graph that should not move during
+        the animation. By default, the first entry is pinned to the origin
```
Legersky, line 633: This is not `animate` method...

Legersky (commented):
Please, could you keep the option to animate using SVG, for instance as method `animate2D_svg`?

pyrigi/framework.py
matthiashimmelmann: ✅

pyrigi/motion.py
matthiashimmelmann: ✅

pyrigi/motion.py
matthiashimmelmann: ✅

pyrigi/motion.py
matthiashimmelmann: ✅

pyrigi/motion.py
```diff
@@ -40,186 +51,380 @@ def graph(self) -> Graph:
...
+            (max_val - min_val) * aspect_ratio[0],
+            (max_val - min_val) * aspect_ratio[1],
+            z_width=(max_val - min_val) * aspect_ratio[2],
+            padding=plot_style.padding,
+        )
 
-            v_dist = np.linalg.norm(r[v2])
-            theta = np.arccos(
-                np.dot([v_dist * t for t in fixed_direction], r[v2]) / v_dist**2
+        from pyrigi import _plot
```
Legersky, line 174: I think here it can be imported above without having a circular import.

pyrigi/motion.py
```diff
@@ -40,186 +51,380 @@ def graph(self) -> Graph:
...
+    def animate2D_plt(
+        self,
         realizations: Sequence[dict[Vertex, Point]],
-        width: int,
-        height: int,
-        spacing: int,
-    ) -> list[dict[Vertex, Point]]:
+        plot_style: PlotStyle,
+        edge_coloring: Sequence[Sequence[Edge]] | dict[str, Sequence[Edge]] = None,
+        delay: int = 50,
```
Legersky, line 261: Could you add the option `duration` similarly to `animate2D_svg`?

pyrigi/motion.py
```diff
@@ -295,6 +501,44 @@ def animate(
                 file.write(svg)
         return SVG(data=svg)
 
+    def animate(
+        self,
+        realizations: Sequence[dict[Vertex, Point]],
+        plot_style: PlotStyle,
+        svg: bool = True,
```
Legersky, line 508: Should we change it to `format: Literal["svg", "matplotlib"] = "svg"` to keep the option to add other formats in the future?

pyrigi/motion.py
```diff
@@ -515,6 +758,14 @@ class ApproximateMotion(Motion):
         distance is at least ``turning_threshold`` times as large as the distance
         of the negative infinitesimal flex, then the latter one is chosen instead.
         If instead the animation is too slow, consider increasing this value.
+    fixed_edge:
```
Legersky, line 761: Since according to the example in the tutorial one can fix a non-edge, which I like, maybe we should change the name to `fixed_vertices`?

pyrigi/motion.py
```diff
@@ -515,6 +758,14 @@ class ApproximateMotion(Motion):
...
+        The edge of the underlying graph that is fixed in the list of realizations.
+        By default, the first entry is pinned to the origin
+        and the second is pinned to the `x`-axis.
+    fixed_direction:
+        Vector to which the first direction is fixed. By default, this is given by
+        the first and second entry.
```
Legersky, line 767: Do you mean the first and second entry of `fixed_edge`?

pyrigi/motion.py
Legersky: or `fixed_pair`?

pyrigi/motion.py
matthiashimmelmann: You can have either `delay` or `duration` I guess. I'll add `duration` to both `animate2D` and `animate3D`

matthiashimmelmann: ✅

pyrigi/motion.py
matthiashimmelmann: ✅

pyrigi/motion.py
Legersky: Sorry, I did not realize that `format` is a built-in name, so maybe `animation_format`?

matthiashimmelmann: ✅

Legersky (approved):

PR #219, opened 2025-01-30
is_sparse/tight now mean (2,3)-sparse/tight

Legersky:


MatteoGallet (commented):
Shall we allow the possibility to select an algorithm for `is_sparse`?

Legersky:
I would say that since this is meant as a shortcut, pebble games are the only reasonable choice. If one wants to compare algorithms, he or she can use `is_kl_sparse`.

MatteoGallet (approved):

PR #227, opened 2025-02-03
Doc compilation test on dev and long cells in tutorials

Legersky:
The doc compilation is now tested also on merge to `dev`. To keep the time reasonable, tutorial cells with long execution time can be skipped. This has effect on local compilation of the documentation, but the tags are removed befofe the online documentation is compiled (we have to double check this once we merge to main).


Legersky:
Skipping the tests does not work on GitHub, I will propose a different solution.

PR #233, opened 2025-02-05
edge_coloring -> edge_colors_custom

Legersky:


matthiashimmelmann:
Would you mind adding a `test_plotstyle` file? I think this morally can be done here. I would like to have a consistency test that checks whether `plot_style` accepts its keywords and changes the default values.

MatteoGallet:
> Would you mind adding a `test_plotstyle` file? I think this morally can be done here. I would like to have a consistency test that checks whether `plot_style` accepts its keywords and changes the default values.

I would suggest to open a new PR for that, and just merge this "atomic" one as it is.

PR #256, opened 2025-02-26
General approximate motions

matthiashimmelmann:
+ Added randomization of equation to deal with stressed frameworks
+ Added a least squares approach for selecting the next infinitesimal flex

### DIFF
--- a/pyrigi/framework.py
+++ b/pyrigi/framework.py
@@ -1631,8 +1631,6 @@ class Framework(object):
         vertex_order:
             A list of vertices, providing the ordering for the entries
             of the infinitesimal flexes.
-        numerical:
-            Whether the check is symbolic (default) or numerical.
 
         Examples
         --------
@@ -1683,7 +1681,7 @@ class Framework(object):
         """
         Return non-trivial infinitesimal flexes.
 
-        See :meth:`~Framework.trivial_inf_flexes`.
+        See :meth:`~Framework.inf_flexes` for possible keywords.
 
         Definitions
         -----------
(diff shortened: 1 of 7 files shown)


pyrigi/framework.py
```diff
@@ -1630,6 +1631,8 @@ def trivial_inf_flexes(self, vertex_order: Sequence[Vertex] = None) -> list[Matr
         vertex_order:
             A list of vertices, providing the ordering for the entries
             of the infinitesimal flexes.
+        numerical:
+            Whether the check is symbolic (default) or numerical.
```
Legersky, line 1635: This parameter is not in the signature of the method...

pyrigi/framework.py
```diff
@@ -1676,23 +1679,16 @@ def trivial_inf_flexes(self, vertex_order: Sequence[Vertex] = None) -> list[Matr
...
     @doc_category("Infinitesimal rigidity")
-    def nontrivial_inf_flexes(
-        self, vertex_order: Sequence[Vertex] = None
-    ) -> list[Matrix]:
+    def nontrivial_inf_flexes(self, **kwargs) -> list[Matrix]:
         """
         Return non-trivial infinitesimal flexes.
 
+        See :meth:`~Framework.trivial_inf_flexes`.
+
```
Legersky, line 1687: I would prefer to have it the other way around since I think one is interested in nontrivial flexes more often than in trivial ones.

pyrigi/framework.py
matthiashimmelmann: ✅

pyrigi/framework.py
matthiashimmelmann: ✅ 
I think the reference was just to an incorrect method. Fixed it.

MatteoGallet (approved):

PR #262, opened 2025-03-06
Refactor `test_motion`

matthiashimmelmann:
+ Parametrized some tests in `test_motion`


Legersky:
I had to resolve conflicts for `feature-general-approximate-motions`, so it might be easier to just rebase the new commits on `dev` once #256 is merged.

PR #269, opened 2025-03-07
Added PEP-conforming `__repr__` methods

matthiashimmelmann:


pyrigi/motion.py
```diff
@@ -723,6 +731,16 @@ def __init__(
         if not self.check_edge_lengths():
             raise ValueError("The given motion does not preserve edge lengths!")
 
+    def interval(self) -> tuple[Number]:
+        """Return the underlying interval of the ParametricMotion."""
+        return self._interval
+
+    def parametrization(self, as_points: bool = False) -> dict[Vertex, Point]:
+        """Return a parametrization of the ParametricMotion."""
```
MatteoGallet, line 739: Shall we change `the parametrization` instead of `a parametrization`?

pyrigi/graph.py
```diff
@@ -99,7 +99,9 @@ def __repr__(self) -> str:
         """
         Return a representation.
```
MatteoGallet, line 100: Shall we write `Return a representation of the graph.`?

pyrigi/framework.py
```diff
@@ -136,8 +136,10 @@ def __str__(self) -> str:
         )
 
     def __repr__(self) -> str:
-        """Return the representation"""
-        return self.__str__()
+        """Return a representation"""
```
MatteoGallet, line 139: Shall we write `Return a representation of the framework.`?

pyrigi/graph.py
matthiashimmelmann: ✅

pyrigi/motion.py
matthiashimmelmann: ✅

pyrigi/framework.py
matthiashimmelmann: ✅

pyrigi/motion.py
```diff
@@ -723,6 +731,16 @@ def __init__(
         if not self.check_edge_lengths():
             raise ValueError("The given motion does not preserve edge lengths!")
 
+    def interval(self) -> tuple[Number]:
+        """Return the underlying interval of a ParametricMotion."""
+        return self._interval
```
Legersky, line 736: Just a copy of the attribute should be returned to prevent accidental changes.

matthiashimmelmann: But tuples are immutable. I thought that was the distinction in attribute getters and why we return `self._dim` instead of `deepcopy(self._dim)`?

Legersky: dim is just a number so it is returned by value, contrary to lists whose items can be mutated. You are right that once we retype it to be tuple, we can return it. Just now it is tuple only by type hinting, so if a list is provided, the list is stored.

matthiashimmelmann: ✅

Legersky:
There is still a problem with realizations and parametrizations: currently, some of the returned `__repr__` strings are not valid Python strings, or are parsed differently. Namely, is a realization contains for instance fractions, they would be parsed as floats. Similarly, a parametrization of a motion yields an error. Hence, these have to be returned in `__repr__` with coordinates as strings to be sympified on input. I will set up some test illustrating the problem.

Legersky (approved):

PR #272, opened 2025-03-10
Updated to Python 3.11 in GitHub workflows.

MatteoGallet:


Legersky (approved):

matthiashimmelmann:
Should we mention in the documentation that we require `Make sure that your Python version is at least 3.11.` in the Installation Guide and in Getting Started?

PR #278, opened 2025-03-11
Slight improvement in symbolic zero recognition via SymPy.

MatteoGallet:


matthiashimmelmann:
Can you add the tests that previously failed to `test_framework.py` (even though they may fall into `long_local`)?

PR #281, opened 2025-03-12
Added `ValueError` if `sp.simplify` does not yield a meaningful result

matthiashimmelmann:


pyrigi/misc.py
```diff
@@ -169,7 +169,12 @@ def is_zero(expr, numerical: bool = False, tolerance: float = 1e-9) -> bool:
         The tolerance that is used in the numerical check coordinate-wise.
     """
     if not numerical:
-        return sp.cancel(sp.sympify(expr)).equals(0)
+        zero_bool = sp.simplify(sp.cancel(sp.sympify(expr))).equals(0)
+        if zero_bool is None:
+            raise ValueError(
+                f"The expression `{expr}` could not be simplified by sympy."
```
Legersky, line 175: Shall we rather return `RuntimeError` as this is something that we hope that should not happen? Maybe also ask in the error message to report it as an issue?

Legersky:
Could you replace all occurrences of `.is_zero` by `is_zero()` also in tests?

Legersky (approved):

PR #309, opened 2025-03-23
Feature: Added stable separating set search on flexible graphs

Lastaapps:
In this PR I implemented the Algorithm 1 from [Katie Clinch, Dániel Garamvölgyi, John Haslegrave, Tony Huynh, Jan Legerský, and Anthony Nixon. Stable cuts, NAC-colourings and flexible realisations of graphs. 2024](https://arxiv.org/abs/2412.16018) that can find a stable cut in a flexible graph.

The more interesting part of the PR is related to proposal how can functions be included to the `Graph` class while being in a different module. My proposal is to use `proxy_call` decorator (see changes in `Graph.py`). Documentation generation works just fine, and autocomplete works, at least for me - I use `pyright`. The goal of this decorator design is to prevent doc strings, interface duplication and simple inclusion into the `Graph` class. If you agree this is a good idea how this issue should be handled, please try if autocomplete works with your language server/IDE/... If you don't agree, propose a better design, please.

Also, please add people responsible for PyRigi design to this thread, thanks.


doc/refs.bib
Legersky, line 1: Please, change only items related to the implemented feature.

Legersky (commented):
Please, order the doc string sections according to https://pyrigi.github.io/PyRigi/development/howto/documentation.html#docstrings 

The short description should be in imperative.

pyrigi/_cuts.py
```diff
@@ -0,0 +1,269 @@
...
+    Checks if the given set of vertices is stable in the given graph.
+
+    Definitions
+    -----------
+    :prf:ref:`Stable set <def-stable-set>`
+
+    Parameters
+    ----------
+    graph:
+        Vertexhe graph to check
```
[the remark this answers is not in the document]
Lastaapps: It is a side effect of me removing generics improperly by doing a replacement `T -> Vertex`. I fixed the cases where `True` got turned into `Vertexrue`, I missed this one. Thanks

doc/refs.bib
Lastaapps: Sorry, auto-formatter got me.

pyrigi/_cuts.py
```diff
@@ -0,0 +1,272 @@
...
+    vertices = _to_vertices(vertices)
+    from pyrigi import Graph as PRGraph
+
+    PRGraph._input_check_vertex_members(graph, vertices)
+
+    if copy:
+        graph = nx.Graph(graph)
+
+    graph.remove_nodes_from(vertices)
+    return not nx.is_connected(graph)
```
Legersky, line 127: If `copy=False`, then the vertices should be added back after removal, in order not to change the original graph.

Legersky: Same for `is_separating_set_dividing`.

pyrigi/graph.py
```diff
@@ -3525,34 +3559,44 @@ def intersection(self, other_graph: Graph) -> Graph:
...
+
+    @doc_category("Generic rigidity")
+    @proxy_call(stable_cut_in_flexible_graph)
+    def stable_cut_in_flexible_graph(self):
+        pass
+
+    @doc_category("Generic rigidity")
+    @proxy_call(stable_cut_in_flexible_graph_fast)
+    def stable_cut_in_flexible_graph_fast(self):
+        pass
```
Legersky, line 3599: The `doc_category` should be `@doc_category("General graph theoretical properties")`.

Legersky (commented):
We should agree on a single term among cut/cutset/separating set and stick to it everywhere.

pyrigi/_cuts.py
Lastaapps: First, I wanted to keep using `nx.Graph` class that can also have an immutable implementation (a graph view). For example, `nx.induced_subgraph` produces such a graph. And for such graphs a copy has to be made as no changes are allowed. I'll check if this mutability information can be obtained from NetworkX, if not, I'll assume PyRigi graph only for now.

pyrigi/_cuts.py
```diff
@@ -0,0 +1,272 @@
...
+    Converts multiple input formats into the graph separator.
+    """
+    if isinstance(vertices, set):
+        return vertices
+    if isinstance(vertices, SeparatingCut):
+        return vertices.cut
+    return set(vertices)
+
+
+def stable_set_violation(
```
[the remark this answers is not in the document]
Legersky: I meant for instance that an obstacle would be returned depending on a parameter, in order to introduce less functions to API.

pyrigi/_cuts.py
Legersky: I don't think this is a problem of (im)mutability. The default behavior should be such that even if the computation fails for some reason (for instance is interrupted by a user), the input graph is preserved, i.e., using copy. If one is aware of the risk that the input graph can change if the computation fails, then it is fine to allow `copy=False` option, but the input graph should be modified back to its original state at the end of the computation.

pyrigi/_flexible_graph_stable_cut.py
```diff
@@ -152,6 +153,11 @@ def stable_cut_in_flexible_graph_fast(
         For a valid input a :class:`pyrigi.data_type.StableCut` in the graph is returned.
         For rigid graphs and cases when ``u`` and ``v`` are not in the same
         rigid component, ``None`` is returned.
+
+    Note
+    ----
+        Vertex and edge metadata may be dropped, if you use them,
+        make copy of your graph first.
```
Legersky, line 160: This cannot be the default behaviour.

pyrigi/_flexible_graph_stable_cut.py
```diff
@@ -0,0 +1,322 @@
...
+
+from pyrigi.data_type import StableCut
+from pyrigi._cuts import stable_set_violation
+import networkx as nx
+import numpy as np
+
+from pyrigi.data_type import Vertex
+
+
+def stable_cutset_in_flexible_graph(
```
Legersky, line 18: Please, rename it to `stable_separating_set` and introduce parameter `check_flexible` instead, similarly to `check_min_rigid` in https://pyrigi.github.io/PyRigi/userguide/api/graph.html#pyrigi.graph.Graph.number_of_realizations

pyrigi/_flexible_graph_stable_cut.py
```diff
@@ -0,0 +1,322 @@
...
+
+    if not isinstance(graph, PRGraph):
+        graph = PRGraph(subgraph)
+
+    cut = _process(graph, u, v)
+    StableCut(a=cut.a, b=cut.b | set(x for c in other for x in c), cut=cut.cut)
+    return cut
+
+
+def stable_cutset_in_flexible_graph_fast(
```
Legersky, line 116: There should not be two methods doing the same thing, we use parameter `algorithm` for such cases, see for instance https://pyrigi.github.io/PyRigi/userguide/api/graph.html#pyrigi.graph.Graph.is_rigid

pyrigi/data_type.py
```diff
@@ -58,3 +59,71 @@
...
+
+    def __repr__(self) -> str:
+        return f"SeparatingCut({self.a}, {self.b} - {self.cut})"
+
+    def __eq__(self, other) -> bool:
+        if self.cut != other.cut:
+            return False
+        if self.a == other.a:
+            return self.b == other.b
+        return self.a == other.b and self.b == other.a
```
Legersky, line 94: Such object does not make sense without a graph. Moreover, mathematically a separating set is just a set of vertices so I do not see a reason to have a special class for it. Similarly for `StableCut` below.

pyrigi/graph.py
```diff
@@ -23,10 +22,46 @@
...
+    stable_cutset_in_flexible_graph_fast,
+)
+from pyrigi._cuts import (
+    stable_set_violation,
+    is_stable_set,
+    is_separating_set,
+    is_separating_set_dividing,
+    is_stable_cutset,
+    is_stable_cutset_dividing,
+)
```
Legersky, line 36: Since these are used just once, I would suggest to just use full paths when they are needed.

pyrigi/_cuts.py
```diff
@@ -0,0 +1,304 @@
...
+
+    vertices = _to_vertices(vertices)
+    from pyrigi import Graph as PRGraph
+
+    PRGraph._input_check_vertex_members(graph, vertices)
+
+    return _revertable_set_removal(graph, vertices, lambda g: not nx.is_connected(g))
+
+
+def is_separating_set_dividing(
```
Legersky, line 168: We should find a better name with corresponding math terminology: for instance $(u,v)$-separating set (`is_uv_separating_set`), or $S$ separates vertices $u$ and $v$ (e.g. `separates_two_vertices`)

pyrigi/_cuts.py
```diff
@@ -0,0 +1,304 @@
...
+
+    Note
+    ----
+        See :meth:`~pyrigi.graph.Graph.is_stable_set` and
+        :meth:`~pyrigi.graph.Graph.is_separator`.
+    """
+    return is_stable_set(graph, vertices) and is_separating_set(graph, vertices)
+
+
+def is_stable_cutset_dividing(
```
Legersky, line 262: I think this could be omitted as one can easily check whether a set is stable and separates $u$ and $v$.

pyrigi/_cuts.py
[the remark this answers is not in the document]
Lastaapps: I switched to a tuple `(verdict, Option[counterexample])`

doc/math/graph_theory.md
```diff
@@ -121,6 +121,39 @@ $G-U$ is not {prf:ref}`connected <def-connected>`.
...
+
+{{pyrigi_crossref}} {meth}`~.Graph.is_separating_set_dividing`
+:::
+
+
+:::{prf:definition} stable set
+:label: def-stable-set
+
+The set $S \subset V$ is called a _stable set_ of $G$
+if it is an independent set in $G$.
```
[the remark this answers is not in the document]
Lastaapps: :heavy_check_mark:

MatteoGallet (commented):
Could you please fix the linting error, so that we see the outcome of the tests?

pyrigi/_cuts.py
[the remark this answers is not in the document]
Legersky: I would prefer that the default behavior is that just boolean is output to be consistent with the rest of the package. When `certificate: bool = False`, the output is `(verdict, Option[counterexample])`.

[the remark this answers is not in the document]
Lastaapps: Sure...

pyrigi/_wrap.py
```diff
@@ -0,0 +1,167 @@
...
+import functools
+from inspect import Signature, _empty
+from typing import Any, Callable, TypeVar, cast
+import sys
+from types import GetSetDescriptorType, ModuleType
+
+if sys.version_info >= (3, 10):
+    from typing import ParamSpec
+else:
+    from typing_extensions import ParamSpec
```
MatteoGallet, line 12: Does this check whether we are using Python >= 3.10, right? If so, please just use

```python
from typing import ParamSpec
```

since we globally request Python >= 3.11

pyrigi/_cuts.py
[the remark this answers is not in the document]
Lastaapps: :heavy_check_mark:

test/test_stablecut.py
```diff
@@ -30,7 +29,23 @@ def test_stable_set_violation():
...
+        pass
+
+    _revertable_set_removal(graph2, set(), noop)
+    assert nx.is_isomorphic(graph1, graph2)
+    _revertable_set_removal(graph2, {2}, noop)
+    assert nx.is_isomorphic(graph1, graph2)
+    _revertable_set_removal(graph2, {0, 1, 2, 3}, noop)
+    assert nx.is_isomorphic(graph1, graph2)
+    _revertable_set_removal(nx.induced_subgraph(graph2, [0, 1, 2]), {1}, noop)
+    assert nx.is_isomorphic(graph1, graph2)
```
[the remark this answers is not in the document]
Lastaapps: :heavy_check_mark:

pyrigi/__init__.py
```diff
@@ -5,6 +5,15 @@
...
 """
 
+# Do not export internal functions yet!
+# Internal functions are not exported until the graph's monolit design is decomposed.
+# To use pyrigi.Graph class safely in external functions, typing has to be used.
+# As these functions are referenced also in the Graph clss itself, a dependency
+# cycles is formed. Therefore, (probably) the best approach is to use
+# nx.Graph in external functions for now and call Graph's methods
+# in unsafe way till the monolit is decomposed and we can switch
+# directly to calling new functions (previous methods) directly.
```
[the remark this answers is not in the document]
Lastaapps: :heavy_check_mark:

doc/refs.bib
```diff
@@ -29,6 +29,17 @@ @article{
...
+@misc{
+  ClinchGaramvölgyiEtAl2024,
+  author = {Katie Clinch and Dániel Garamvölgyi and John Haslegrave and Tony Huynh and Jan Legerský and Anthony Nixon},
+  title = {Stable cuts, {{NAC}}-colourings and flexible realisations of graphs},
+  year = {2024},
+  eprint = {2412.16018},
+  archivePrefix = {arXiv},
+  primaryClass = {math.CO},
+  url = {https://arxiv.org/abs/2412.16018},
+}
```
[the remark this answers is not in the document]
Lastaapps: :heavy_check_mark:

pyrigi/data_type.py
Lastaapps: :heavy_check_mark:

pyrigi/graph.py
Lastaapps: :heavy_check_mark:

pyrigi/_cuts.py
Lastaapps: :heavy_check_mark:

pyrigi/_cuts.py
Lastaapps: I guess...

:heavy_check_mark:

pyrigi/_cuts.py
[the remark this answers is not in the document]
Lastaapps: :heavy_check_mark:

pyrigi/_flexible_graph_stable_cut.py
Lastaapps: As the same algorithm is used under the hood, I suggest using parameters setting assumptions about the input graph:
```python
def stable_separating_set(
    graph: nx.Graph,
    check_flexible: TODO,
    u: Optional[Vertex] = None,
    v: Optional[Vertex] = None,
    assume_connected: bool = False,
    assume_distinct_rigid_components: bool = False,
) -> Optional[set[Vertex]]:
```

pyrigi/_flexible_graph_stable_cut.py
Lastaapps: I don't think I understand what the expected behavior should be: You mean, an exception will be thrown for flexible graphs instead of the function returning `None` in it's `Optional` return type?

Legersky: I mean to add `check_flexible=True` to the signature and at the beginning of the function::
```py
if check_flexible and graph.is_rigid(dim=2):
    raise ValueError("The graph must be 2-flexible!")
```

test/test_stablecut.py
Legersky, line 1: please, rename to `test_separating_set.py`

pyrigi/separating_set.py
```diff
@@ -0,0 +1,495 @@
...
+        rigid component, ``None`` is returned.
+    """
+    from pyrigi import Graph as PRGraph
+
+    if check_flexible and graph.is_rigid(dim=2):
+        raise ValueError("The graph must be 2-flexible!")
+
+    # find the smallest connected component and put it first
+    if not check_connected:
+        is_connected = True
```
Legersky, line 304: If I am not mistaken, `connected_components` should be initiated also in this case.

pyrigi/separating_set.py
```diff
@@ -0,0 +1,495 @@
...
+        if v in disallowed:
+            logging.warning(f"Both vertices {u} and {v} share the same rigid component")
+            return None
+    else:
+        # choose a vertex at random
+        v = next(x for x in graph.nodes if x not in disallowed)
+    return v
+
+
+def _process(
```
Legersky, line 403: Could the name be more descriptive, e.g. `_find_stable_uv_separating_set`?

pyrigi/separating_set.py
```diff
@@ -0,0 +1,495 @@
...
+                raise ValueError(
+                    "Chosen vertices must not be in the same rigid component"
+                )
+            case new_v:
+                v = new_v
+
+    return _process(graph, u, v)
+
+
+def _find_and_validate_u_and_v(
```
Legersky, line 359: Could the name be more descriptive, e.g., `_validate_uv_different_rigid_comps`?

test/test_stablecut.py
Lastaapps: :heavy_check_mark:

pyrigi/separating_set.py
Lastaapps: :heavy_check_mark:

pyrigi/separating_set.py
Lastaapps: :heavy_check_mark:

Lastaapps: :heavy_check_mark:

pyrigi/separating_set.py
Lastaapps: :heavy_check_mark:

pyrigi/separating_set.py
Lastaapps: :heavy_check_mark:

test/test_stablecut.py
Lastaapps: :heavy_check_mark:

pyrigi/separating_set.py
```diff
@@ -0,0 +1,515 @@
...
+    ):
+        """
+        Restore the contracted graph to its original form.
+        Inverse operation for contract.
+        """
+        for n in x_neigh - u_neigh - {u}:
+            graph.remove_edge(u, n)
+        graph.add_node(x, **vertex_data[x])
+        for n in x_neigh:
+            graph.add_edge(x, n, **edge_data[frozenset((n, x))])
```
Legersky, line 489: Please, avoid using `n` for a vertex asi it is deserved for the number of vertices, see https://pyrigi.github.io/PyRigi/development/howto/notation.html

Lastaapps: :heavy_check_mark:

pyrigi/separating_set.py
```diff
@@ -0,0 +1,550 @@
...
+T = TypeVar("T")
+
+
+def is_stable_set(
+    graph: nx.Graph,
+    vertices: Collection[Vertex],
+    certificate: bool = False,
+) -> bool | tuple[bool, Optional[Edge]]:
+    """
+    Return if given ``vertices`` form a stable set.
```
MatteoGallet, line 36: With `is_` methods, we usually write "Return whether" instead of "Return if". Could you please change this?

pyrigi/separating_set.py
```diff
@@ -0,0 +1,550 @@
...
+    Definitions
+    -----------
+    :prf:ref:`Stable set <def-stable-set>`
+
+    Parameters
+    ----------
+    vertices:
+        A set of vertices to be checked.
+    certificate:
+        If ``False``, just a boolean whether the set is stable or not is returned.
```
MatteoGallet, line 50: I would suggest to change

```
just a boolean whether the set is stable or not is returned
```

by

```
a boolean is returned whether the set is stable or not
```

pyrigi/separating_set.py
```diff
@@ -0,0 +1,550 @@
...
+                graph.add_edge(u, v, **edge_data[(u, v)])
+
+
+def is_separating_set(
+    graph: nx.Graph,
+    vertices: Collection[Vertex],
+    use_copy: bool = True,
+) -> bool:
+    """
+    Return if ``vertices`` are a separating set.
```
MatteoGallet, line 135: With is_ methods, we usually write "Return whether" instead of "Return if". Could you please change this?

pyrigi/separating_set.py
```diff
@@ -0,0 +1,550 @@
...
+
+def is_uv_separating_set(
+    graph: nx.Graph,
+    vertices: Collection[Vertex],
+    u: Vertex,
+    v: Vertex,
+    use_copy: bool = True,
+) -> bool:
+    """
+    Return if ``vertices`` separate the vertices ``u`` and ``v``.
```
MatteoGallet, line 187: With is_ methods, we usually write "Return whether" instead of "Return if". Could you please change this?

pyrigi/separating_set.py
```diff
@@ -0,0 +1,550 @@
...
+    return _revertable_set_removal(graph, vertices, check_graph, use_copy=use_copy)
+
+
+def is_stable_separating_set(
+    graph: nx.Graph,
+    vertices: Collection[Vertex],
+    use_copy: bool = True,
+) -> bool:
+    """
+    Return if ``vertices`` are a stable separating set.
```
MatteoGallet, line 243: With is_ methods, we usually write "Return whether" instead of "Return if". Could you please change this?

pyrigi/separating_set.py
```diff
@@ -0,0 +1,552 @@
...
+    Definitions
+    -----------
+    :prf:ref:`Separating set <def-separating-set>`
+
+    Parameters
+    ----------
+    vertices:
+        The vertices to check.
+    use_copy:
+        Create a copy of the graph before the vertices are removed
```
MatteoGallet, line 148: I would suggest to add `If ``True`` (default), ...`

pyrigi/separating_set.py
```diff
@@ -0,0 +1,552 @@
...
+    Parameters
+    ----------
+    vertices:
+        Vertex set to remove.
+    func:
+        A function whose result is returned for the graph with vertices removed.
+        Warning: if ``func`` modifies its input graph, then it is not guaranteed
+        that the ``graph`` is restored correctly by ``_revertable_set_removal``.
+    use_copy:
+        Create a copy of the graph before the vertices are removed
```
MatteoGallet, line 102: I would suggest to add `If ``True`` (default), ...`

pyrigi/separating_set.py
```diff
@@ -0,0 +1,552 @@
...
+
+    Parameters
+    ----------
+    vertices:
+        The set of vertices to be checked to separate ``u`` and ``v``.
+        If ``u`` or ``v`` is contained in ``vertices``,
+        ``ValueError`` is raised.
+    u, v:
+    use_copy:
+        Create a copy of the graph before the vertices are removed
```
MatteoGallet, line 203: I would suggest to add `If ``True`` (default), ...`

pyrigi/separating_set.py
```diff
@@ -0,0 +1,552 @@
...
+    --------
+    >>> import pyrigi.graphDB as graphs
+    >>> H = graphs.Cycle(5)
+    >>> H.is_uv_separating_set([1,3], 0, 2)
+    True
+    >>> H.is_uv_separating_set([2,4], 0, 1)
+    False
+    """
+    if u in vertices:
+        raise ValueError(f"u={u} is in the separating set")
```
MatteoGallet, line 217: I would write a different error message, since the point here seems to me that `u` should not be among the elements in `vertices`. Hence, I would suggest to write:

```
f"The vertex u={u} should not be among the ones in the parameter `vertices`"
```

pyrigi/separating_set.py
```diff
@@ -0,0 +1,552 @@
...
+    if v in vertices:
+        raise ValueError(f"v={v} is in the separating set")
```
MatteoGallet, line 219: I would write a different error message, since the point here seems to me that `b` should not be among the elements in `vertices`. Hence, I would suggest to write:

```
f"The vertex v={v} should not be among the ones in the parameter `vertices`"
```

pyrigi/separating_set.py
```diff
@@ -0,0 +1,552 @@
...
+
+    Definitions
+    -----------
+    * :prf:ref:`Stable set <def-stable-set>`
+    * :prf:ref:`Separating set <def-separating-set>`
+
+    Parameters
+    ----------
+    vertices:
+    use_copy:
```
MatteoGallet, line 259: Shall we remove these lines?

pyrigi/separating_set.py
```diff
@@ -0,0 +1,552 @@
...
+    Find a stable separating set if the graph is 2-flexible.
+
+    Algorithm 1 in :cite:p:`ClinchGaramvölgyiEtAl2024` is used.
+    The algorithm allows to specify two vertices ``u`` and ``v``
+    that are separated by the returned stable separating set,
+    provided that they are not in the same
+    :prf:ref:`2-rigid component <def-rigid-components>`.
+    Alternatively, a single vertex ``u`` can be specified that is
+    avoided in the returned stable separating set,
+    provided it is not contained in all 2-rigid components.
```
MatteoGallet, line 293: I would suggest to put this part under the "Notes" section in the docstring.

pyrigi/separating_set.py
```diff
@@ -0,0 +1,552 @@
...
+        pyrigi._graph_input_check.vertex_members(graph, v)
+
+    ############################################################################
+    # Connectivity
+    ############################################################################
+
+    # if the graph is not connected, we can possibly reduce the work needed
+    # by finding a connected component that contains u
+    # and find a separating set in it or just calling it the day
+    # if v is not specified or lies in another component
```
MatteoGallet, line 355: I would suggest to slightly rephrase:

```
    # and find a separating set in it or return an empty set
    # if v is not specified or lies in another component
```

pyrigi/separating_set.py
Legersky: No, since if we do that, then the doc does not render correctly.

pyrigi/separating_set.py
Legersky: :heavy_check_mark:

pyrigi/separating_set.py
Legersky: :heavy_check_mark:

pyrigi/separating_set.py
Legersky: :heavy_check_mark:

pyrigi/separating_set.py
Legersky: :heavy_check_mark:

pyrigi/separating_set.py
Legersky: :heavy_check_mark:

pyrigi/separating_set.py
Legersky: :heavy_check_mark:

pyrigi/separating_set.py
Legersky: :heavy_check_mark:

pyrigi/separating_set.py
Legersky: :heavy_check_mark:

pyrigi/separating_set.py
Legersky: :heavy_check_mark:

doc/math/graph_theory.md
```diff
@@ -126,12 +126,28 @@ where two paths are internally disjoint if they do not share any edge.
...
+if $u$ and $v$ are in different connected components of $G-U$. 
 
 {{pyrigi_crossref}} {meth}`~.Graph.is_separating_set`
+{meth}`~.Graph.is_uv_separating_set`
+:::
+
+
+:::{prf:definition} stable set
+:label: def-stable-set
+
```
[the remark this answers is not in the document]
Legersky: :heavy_check_mark:

pyrigi/separating_set.py
```diff
@@ -0,0 +1,552 @@
...
+                    return False, (v, u)
+                else:
+                    return False
+    if certificate:
+        return True, None
+    else:
+        return True
+
+
+def _revertable_set_removal(
```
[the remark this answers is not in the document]
Legersky: `_remove_apply_restore`?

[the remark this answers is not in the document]
MatteoGallet: I like that!

MatteoGallet (approved):

PR #315, opened 2025-03-25
Added the option to save a figure or an animation to a file

matthiashimmelmann:


pyrigi/motion.py
```diff
@@ -176,6 +178,11 @@ def animate3D(
             The omitted edges are given the value ``plot_style.edge_color``.
         duration:
             The duration of one period of the animation in seconds.
+        save_animation:
+            A boolean indicating whether the figure is saved as a `.gif` file.
```
MatteoGallet, line 182: Shall we add

```
If ``True``, the `save` method from `matplotlib` is invoked, which uses external writers to create the GIF file, such as `ffmpeg` (default) or `pillow`.
```

?

Legersky:
Shall we use the same logic as was in `animate2D_svg`? Namely, if `filename` is provided, image/animation will be stored. The currently proposed implementation changes the behavior of `animate2D_svg` and requires two parameter to be specified, unless the default name is used.

MatteoGallet (approved):

PR #317, opened 2025-03-26
Added `inplace` parameter to the methods `is_min_inf_rigid` and `is_redundantly_inf_rigid`

matthiashimmelmann:


pyrigi/framework.py
```diff
@@ -2051,6 +2051,13 @@ def is_min_inf_rigid(self, **kwargs) -> bool:
         -----
         :prf:ref:`Minimal infinitesimal rigidity <def-min-rigid-framework>`
 
+        Parameters
+        ----------
+        inplace:
+            If ``True``, the framework is modified,
```
MatteoGallet, line 2057: Shall we rather say something along the line `the framework is modified during the algorithm and later restored to its initial status`?

matthiashimmelmann: ✅

Legersky:
I would suggest to use different parameter name than `inplace` (for instance `use_copy`), since `inplace` is used in `zero_extension` etc in slightly different meaning: there the output is actually a modified graph and the parameter indicates whether it is a new one, or the changes are applied to the existing one.

matthiashimmelmann:
> I would suggest to use different parameter name than `inplace` (for instance `use_copy`), since `inplace` is used in `zero_extension` etc in slightly different meaning: there the output is actually a modified graph and the parameter indicates whether it is a new one, or the changes are applied to the existing one.

What about `rotate2D` etc.?

Legersky:
There `inplace` has the same meaning as in `zero_extension`.

Legersky (approved):

PR #327, opened 2025-04-09
Feature: Added the method `rotate3D` and the alias `rotate`

matthiashimmelmann:
+ Methods added
+ Corresponding tests added

### DIFF
--- a/pyrigi/framework.py
+++ b/pyrigi/framework.py
@@ -2727,6 +2727,7 @@ class Framework(object):
             Otherwise, a new rotated framework is returned.
         """
         _input_check.dimension_for_algorithm(self.dim, [2], "rotate2D")
+        _input_check.equal(len(point_axis), 2, "length of the `point_axis`")
 
         rotation_matrix = Matrix(
             [[sp.cos(angle), -sp.sin(angle)], [sp.sin(angle), sp.cos(angle)]]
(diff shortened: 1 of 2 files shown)


Legersky:
`point_axis` and `direction_axis` sound a bit strange to me, maybe `axis_shift` and `axis_direction`  instead?
And `rotation_center` for `rotate2D`?

matthiashimmelmann:
> `point_axis` and `direction_axis` sound a bit strange to me, maybe `axis_shift` and `axis_direction` instead? And `rotation_center` for `rotate2D`?

I guess the natural, mathematical concept would be `rotation_axis` or `axis_of_rotation`. Why don't we use that instead? `rotation_center` works for me.

MatteoGallet:
> > `point_axis` and `direction_axis` sound a bit strange to me, maybe `axis_shift` and `axis_direction` instead? And `rotation_center` for `rotate2D`?
> 
> I guess the natural, mathematical concept would be `rotation_axis` or `axis_of_rotation`. Why don't we use that instead? `rotation_center` works for me.

The problem for me regarding `rotation_axis` or `axis_of_rotation` is that I think of that, in general, as any axis in space, not just an axis through the origin. Hence, for me, if we have a parameter `rotation_axis`, it should be constituted of, say, the 6 Pluecker coordinates of the axis, not just the 3 coordinates of its direction. That is the reason why I proposed a different name for the direction.

Legersky (approved):

PR #330, opened 2025-04-15
Code: Moved sparsity and generic rigidity implementations to separate modules

Legersky:
The implementations of methods for generic rigidity and sparsity have been moved to separate modules `generic_rigidity.py` and `sparsity.py`. Corresponding tests have been moved as well and adjusted to test the wrapped functions as well (this can be switched off by setting `test_graph.TEST_WRAPPED_FUNCTIONS = False`.

This required some further adjustments:
- introduced `_get_pebble_digraph` function
- allowed a wrapped function to return `nx.Graph` while the wrapping method `pyrigi.Graph`
- allowed `nx.Graph` as input of `Framework` (followed by retyping it to `Graph`)
- changed classmethod `_warn_randomized_alg` to a function in `warning.py`

### DIFF
--- a/pyrigi/conftest.py
+++ b/pyrigi/conftest.py
@@ -1,5 +1,5 @@
 """
-This file allows to skip `from pyrigi import Graph`
+This file allows skipping `from pyrigi import Graph`
 in section Examples in docstrings.
 """
 
--- a/pyrigi/sparsity.py
+++ b/pyrigi/sparsity.py
@@ -1,5 +1,5 @@
 """
-This module provides algorithms related generic rigidity.
+This module provides algorithms related to graph sparsity.
 """
 
 import math


pyrigi/graph.py
```diff
@@ -1905,215 +1704,38 @@ def is_min_k_redundantly_rigid(
...
+        return pyrigi.generic_rigidity.is_rigid(
+            self,
+            dim=dim,
+            algorithm=algorithm,
+            use_precomputed_pebble_digraph=use_precomputed_pebble_digraph,
+            prob=prob,
+        )
 
     @doc_category("Generic rigidity")
+    @copy_doc(pyrigi.generic_rigidity.is_rigid)
```
MatteoGallet, line 1724: Shouldn't it be 

```python
@copy_doc(pyrigi.generic_rigidity.is_min_rigid)
```

instead?

test/test_signature.py
```diff
@@ -166,13 +166,18 @@ def _assert_same_sign(method: Callable[..., T], func: Callable[..., T]) -> None:
...
-        )
+        if sgn_method.return_annotation == Graph and issubclass(
+            sgn_method.return_annotation, sgn_func.return_annotation
+        ):
+            pass
+        else:
+            raise TypeError(
+                f"""
+                    Method's return type does not match the one of proxy function
+                    method  [{method.__name__}]={sgn_method.return_annotation}
```
MatteoGallet, line 177: Here I would suggest:

```
The return type of the method does not match the one of the proxy function
```

pyrigi/graph.py
Legersky: :heavy_check_mark:

test/test_signature.py
Legersky: :heavy_check_mark:

MatteoGallet (approved):

PR #341, opened 2025-05-16
Code: Split ApproximateMotion and ParametricMotion out of `motion.py`

MatteoGallet:


pyrigi/__init__.py
```diff
@@ -8,5 +8,7 @@
 from pyrigi.framework import Framework
 from pyrigi.graph import Graph
 from pyrigi.graph_drawer import GraphDrawer
-from pyrigi.motion import ApproximateMotion, Motion, ParametricMotion
+from pyrigi.motion import Motion
+from pyrigi.motion.approximate_motion import ApproximateMotion
+from pyrigi.motion.parametric_motion import ParametricMotion
```
Legersky, line 13: The original way should works as well (otherwise we have backward-compatibility issues).

test/motion/test_approximate_motion.py
Legersky, line 1: `ApproximateMotion` could be dropped from test names as it is now clear from the filename.
Similarly for with `ParametricMotion`.

test/motion/test_motion.py
```diff
@@ -1,14 +1,8 @@
...
-
-import numpy as np
 import pytest
-import sympy as sp
 
 import pyrigi.frameworkDB as fws
 import pyrigi.graphDB as graphs
-from pyrigi import ApproximateMotion, Framework, Graph, Motion, ParametricMotion
-from pyrigi.misc.misc import is_zero_vector
+from pyrigi import ApproximateMotion, Graph, Motion, ParametricMotion  # noqa: F401
```
Legersky, line 5: I think that `Graph` is not used.

MatteoGallet: To be honest, I do not understand why, but if I remove it, I get an error in the tests...

pyrigi/__init__.py
MatteoGallet: :white_check_mark:

test/motion/test_approximate_motion.py
MatteoGallet: :white_check_mark:

test/motion/test_motion.py
Legersky: Sorry, it is used in eval of Motion representation, although it is not directly used in the code.

Legersky (approved):

PR #364, opened 2025-05-29
Doc: Added citing PyRigi

Legersky:


matthiashimmelmann (commented):
Do we need to have the clear-text citation? Maybe we can simply add a link to the preprint below the `.bib` citation á la "A preprint of this paper is [available here](https://doi.org/10.48550/arXiv.2505.22652)."

PR #369, opened 2025-06-11
Hotfix: Newton's method in `ApproximateMotion` not convergent in some cases

matthiashimmelmann:
+ Fixed error where animation went out of frame
+ Added missing feature that lets us only use the fix_edge parameter in the Motion class in 2D
+ Fixed bug that caused problematic behavior in 3D ApproximateMotion.

### DIFF
--- a/pyrigi/motion/approximate_motion.py
+++ b/pyrigi/motion/approximate_motion.py
@@ -11,7 +11,7 @@ import numpy as np
 
 import pyrigi._utils._input_check as _input_check
 from pyrigi._utils._conversion import sympy_expr_to_float
-from pyrigi._utils._zero_check import is_zero
+from pyrigi._utils._zero_check import is_zero, is_zero_vector
 from pyrigi._utils.linear_algebra import _normalize_flex, _vector_distance_pointwise
 from pyrigi.data_type import (
     DirectedEdge,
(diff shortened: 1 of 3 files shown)


pyrigi/motion/motion.py
```diff
@@ -191,18 +191,23 @@ def animate3D(
 
         x_coords, y_coords, z_coords = [
             [
-                realization[v][0]
+                realization[v][i]
```
matthiashimmelmann, line 194: Bug fixed: Now the animation fits in the canvas.

pyrigi/motion/approximate_motion.py
```diff
@@ -629,18 +693,37 @@ def _newton_steps(self, realization: dict[Vertex, Point]) -> dict[Vertex, Point]
...
+        cur_error = prev_error = np.linalg.norm(equations)
+        damping = 0.2
         rand_mat = np.random.rand(
             F._graph.number_of_edges() - self._stress_length, F._graph.number_of_edges()
         )
-        while not cur_error < self.tolerance:
-            rigidity_matrix = np.array(F.rigidity_matrix()).astype(np.float64)
+        while cur_error > self.tolerance:
+            rigidity_matrix = np.array(
+                F.rigidity_matrix(edge_order=self._edge_lengths.keys())
```
matthiashimmelmann, line 717: Bug fixed: edge order prevents convergence

PR #373, opened 2025-07-05
Feature: Export `PlotStyle` and custom colors from `GraphDrawer`

matthiashimmelmann:
Now, we can create a `Framework` or `Graph` from a `GraphDrawer` plot and export the color attributes. It is also possible to export custom color attributes that have been set in the ` GraphDrawer` using the methods `custom_colors`, `vertex_colors_custom` and `edge_colors_custom`. The most recent style choices can be exported as a `PlotStyle2D` via the method `plot_style`.

### DIFF
--- a/pyrigi/graph_drawer/graph_drawer.py
+++ b/pyrigi/graph_drawer/graph_drawer.py
@@ -897,8 +897,15 @@ class GraphDrawer(object):
         Return the plot style that is currently set in the graph drawer.
         """
         vertex_labels = True if (self._vertex_labels == 1) else False
-        return PlotStyle2D(vertex_color=self._vertex_color, edge_color=self._edge_color, vertex_size=20*self._radius, vertex_labels=vertex_labels, edge_width=0.8*self._edge_width, font_size=int(round(0.85*self._label_size)))
-    
+        return PlotStyle2D(
+            vertex_color=self._vertex_color,
+            edge_color=self._edge_color,
+            vertex_size=20 * self._radius,
+            vertex_labels=vertex_labels,
+            edge_width=0.8 * self._edge_width,
+            font_size=int(round(0.85 * self._label_size)),
+        )
+
     def custom_colors(self):
         """
         Return the colors of each vertex and edge in the graph drawer.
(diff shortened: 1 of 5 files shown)


pyrigi/graph_drawer/graph_drawer.py
```diff
@@ -890,3 +891,51 @@ def framework(self, grid: bool = False) -> Framework:
             for v in H.nodes:
                 posdict[v] = [Rational(x, self._grid_size) for x in posdict[v]]
         return Framework(graph=H, realization=posdict)
+
+    def plot_style(self):
+        """
+        Return the plot style that is currently set in the graph drawer.
+        """
+        vertex_labels = True if (self._vertex_labels == 1) else False
```
Legersky: Is there a reason, why this is explicitly compared to `1`?

pyrigi/graph_drawer/graph_drawer.py
```diff
@@ -890,3 +891,51 @@ def framework(self, grid: bool = False) -> Framework:
...
+            edge_width=0.8 * self._edge_width,
+            font_size=int(round(0.85 * self._label_size)),
+        )
+
+    def custom_colors(self):
+        """
+        Return the colors of each vertex and edge in the graph drawer.
+
+        The custom vertex and edge colors are returned as a tuple of two dictionaries.
+        """
```
Legersky: I would keep only the two separate methods for edge colors and vertex colors.

pyrigi/graph_drawer/graph_drawer.py
matthiashimmelmann: No reason. Fixed.

pyrigi/motion/motion.py
```diff
@@ -618,8 +668,11 @@ def animate2D_svg(
     def animate(
         self,
         realizations: Sequence[dict[Vertex, Point]],
-        plot_style: PlotStyle,
-        animation_format: Literal["svg", "matplotlib"] = "svg",
+        vertex_colors_custom: (
+            Sequence[Sequence[Vertex]] | dict[str, Sequence[Vertex]]
+        ) = None,
+        edge_colors_custom: Sequence[Sequence[Edge]] | dict[str, Sequence[Edge]] = None,
+        animation_format: str | Literal["svg", "matplotlib"] = "svg",
```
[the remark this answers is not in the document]
matthiashimmelmann: I think `Literal` is supposed to be more restrictive than `str` by allowing only certain values. I moved this to an explicit input check.

pyrigi/graph_drawer/graph_drawer.py
matthiashimmelmann: This is done implicitly (due to dynamic typing), but this may be unsafe. It is done explicitly now.

Legersky (approved):

PR #376, opened 2025-07-17
Fix Throw Warning when numerical coordinates are detected

matthiashimmelmann:


pyrigi/warning.py
```diff
@@ -104,3 +104,38 @@ def _warn_randomized_alg(
...
+        msg: str = None,
+        class_off: object = None,
+        *args,
+    ):
+        if msg is not None:
+            super().__init__(msg, *args)
+        else:
+            msg_str = (
+                "\nNumerical coordinates were detected in the Framework's realization."
+                + " If symbolic computations are performed on this framework,"
```
Legersky: Should we indicate the coordinates of the framework and in which method this was raised?

PR #393, opened 2026-05-09
Test: Add test_wrapper_parameter_forwarding to replace TEST_WRAPPED_FUNCTIONS

khantnyarhein:
Adds a parametrized test that verifies all @copy_doc wrapper methods on Graph and Framework correctly forward every parameter to their underlying proxy function unchanged.

The test uses unittest.mock.patch to intercept the actual proxy call and asserts both that the graph/framework instance is passed as the first positional argument and that all other parameters are forwarded with their original values. A helper _find_patch_target resolves the correct patch site by walking the method's module namespace.

This makes per-method entries in TEST_WRAPPED_FUNCTIONS redundant.

I will be deduplicating the wrapper-wrapped double assertions later in smaller PRs when the current PR is approved and merged to the epic branch.


khantnyarhein:
- VAR_KEYWORD params: filtered **kwargs out of mock_args; inject a _test_kwarg sentinel instead to actually verify the catch-all is forwarded
- None sentinel: replaced with _MISSING = object() to distinguish a missing key from a key with value None
- Reverse check: added assertion that the proxy receives no arguments beyond what mock_args sent and it catches wrappers that inject extra keyword or positional args
- _assert_name_matches: assertion for wrapper method and wrapped functions to have the same name as the test suggests

Note: the positional fallback was flagged as dead code but is needed: Framework.rescale and Graph.is_critically_k_edge_apex forward a parameter positionally to the proxy.

If the implementation looks good, I will move the current test_signature.py into wrapper folder in test and decouple them into separate more testable and modular pieces.

Legersky:
@khantnyarhein Thanks for the changes, please go ahead with decoupling.

khantnyarhein:
@Legersky I have tried to move the test to test/wrapper folder in commit - https://github.com/PyRigi/PyRigi/commit/d83c66f66fdf78f5c2409172159bb7a747f06a4d

test/wrapper/_helpers.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+def _assert_name_matches(cls: Type, attr_name: str, proxy_func_name: str) -> None:
+    assert attr_name == proxy_func_name, (
+        f"{cls.__name__}.{attr_name} is decorated with @copy_doc({proxy_func_name}) "
+        f"but the method name differs from the proxy function name"
+    )
+
+
+def _assert_params_forwarded(
```
Legersky: If I remember correctly from our discussion, checking that an `AssertionError` was raised in `test_bad_wrapper_detected` was not working well since it can pass due to a different problem. Does it make sense, instead of asserts, to return `(True, "')` when all checks pass and `(False, "problem description")` when something fails, and assert the result in the outer function?

This could also allow to rewrite the `with patch.object(patch_module, patch_name)` blocks in `test_wrapper_parameter_forwarding` and `test_bad_wrapper_detected` so that it would be clear that they test/meta-test the same thing.

test/wrapper/_helpers.py
```diff
@@ -0,0 +1,123 @@
...
+            f"{cls.__name__}.{attr_name} didn't pass "
+            f"the _BadWrappers instance as first arg."
+        )
+    elif cls == Graph:
+        assert isinstance(
+            first_arg, nx.Graph
+        ), f"{cls.__name__}.{attr_name} didn't pass the Graph instance as first arg."
+    elif cls == Framework:
+        assert isinstance(
+            first_arg, Framework
```
Legersky: Should this be `Framework`, or `FrameworkBase`? There is `nx.Graph` on line 77, not `Graph`. I believe these two should be consistent.

test/test_signature.py
```diff
@@ -242,7 +243,7 @@ def is_subtype(sub_type, super_type):
                 )
 
 
-@pytest.mark.parametrize(("cls"), [Graph, Framework])
+@pytest.mark.parametrize(("cls"), [Graph, Framework, _BadWrappers])
```
Legersky: Please, add here just an explanation why we have `_BadWrappers` here: they have the correct signature, but are still wrapping badly.

khantnyarhein:
I've tried to address the three comments above in this commit - [46eeb33](https://github.com/PyRigi/PyRigi/pull/393/commits/46eeb3333d9bfa742e9211f08fca24dff56d9975)

test/wrapper/_check.py
```diff
@@ -0,0 +1,193 @@
...
+    for i in range(1, len(call_args[0])):
+        if i < len(proxy_param_names):
+            forwarded_names.add(proxy_param_names[i])
+
+    extra_positional_count = max(0, len(call_args[0]) - len(proxy_param_names))
+    if extra_positional_count > 0:
+        return False, (
+            f"{cls.__name__}.{attr_name} forwarded {extra_positional_count} "
+            f"unexpected positional argument(s) to the proxy."
+        )
```
Legersky: Please, refactor as we discussed.

khantnyarhein: Hello, I've refactored in this commit - [c07974e](https://github.com/PyRigi/PyRigi/pull/393/commits/c07974e4f53e976a8bdca4e137e6eef6b9882d1c)

Legersky (approved):

PR #394, opened 2026-05-23
Doc: Converted graph submodule docstrings to function-style and updated copy_doc decorator logic

khantnyarhein:
The Graph class exposes methods via @copy_doc. Before this PR, copy_doc simply copied the docstring verbatim from the underlying module-level function, so all wrapped functions had to hold method-style docstrings.

This PR introduces function-style as the canonical docstring format for wrapped functions and updates copy_doc to convert them to method-style at import time via func_to_method_doc.

Changes

- Extended copy_doc in pyrigi/_utils/_doc.py to call func_to_method_doc on the copied docstring, transforming function-style examples (>>> is_rigid(G)) into method-style (>>> G.is_rigid()) for the class API docs.

- Implemented func_to_method_doc and its helpers (_transform_doctest_func_to_method, _transform_doctest_nx_to_graph_alias, _transform_func_to_meth_refs) to handle doctest lines, NetworkX
alias methods, Sphinx :func: references, and "this/the function" text.

- _transform_doctest_func_to_method uses a re.sub fallback (in addition to re.match) so wrapped calls like len(list(all_k_extensions(G, 0))) and type(all_extensions(G)) are also converted correctly.

- Converted docstrings in 11 graph submodule files from method-style to function-style via a one-time migration script (doc/scripts/migrate_docstrings.py).

- Fixed two categories of doctest patterns that initial the migration script(still part of PR)'s regex missed:
  1.  Wrapped calls (len(list(G.method()))) - same re.sub approach applied to the migration helper.
  2.  Chained instantiation (graphs.Diamond().method(...)) - decoupled into an intermediate variable; documented this constraint in the developer guide.

- Added 14 unit tests in test/utils/test__doc.py covering func_to_method_doc and its helpers.

- Added a developer guide note (doc/development/internal_reference/graph/index.md) explaining the two docstring format rules contributors must follow. (this is my initial decision, and can be moved)


doc/development/cheatsheet.md
```diff
@@ -65,10 +65,10 @@ For detailed overview of MyST syntax, see the [MyST documentation](https://myst-
...
 
 :::{csv-table}
-`` :class:`~pyrigi.framework.Framework` `` , {class}`~pyrigi.framework.Framework`
-`` :class:`pyrigi.framework.Framework` `` , {class}`pyrigi.framework.Framework`
-`` :meth:`pyrigi.framework.Framework.delete_edge` `` , {meth}`pyrigi.framework.Framework.delete_edge`
-`` :meth:`~pyrigi.framework.Framework.delete_edge` `` , {meth}`~pyrigi.framework.Framework.delete_edge`
+`` :class:`~pyrigi.framework.framework.Framework` `` , {class}`~pyrigi.framework.framework.Framework`
+`` :class:`pyrigi.framework.framework.Framework` `` , {class}`pyrigi.framework.framework.Framework`
+`` :meth:`pyrigi.framework.framework.Framework.delete_edge` `` , {meth}`pyrigi.framework.framework.Framework.delete_edge`
+`` :meth:`~pyrigi.framework.framework.Framework.delete_edge` `` , {meth}`~pyrigi.framework.framework.Framework.delete_edge`
```
MatteoGallet, line 71: Are these changes related to this PR? If not, I would suggest to revert them.

pyrigi/_utils/_doc.py
```diff
@@ -25,19 +30,221 @@ def copy_doc(
...
-    Copy the docstring from the provided function.
-
-    In tests, it also ensures that the signatures match.
+    Copy the docstring from the provided function, converting it to method-style.
     """
 
     def wrapped(method: Callable[..., T]) -> Callable[P, T]:
-        method.__doc__ = proxy_func.__doc__
+        method_doc = proxy_func.__doc__
+        if method.__qualname__.startswith("Graph."):
```
MatteoGallet, line 38: Is it maybe possible not to hard-code ``Graph`` here, and rather have a dictionary at the beginning of the files where we collect the starting string of each class for which we may want to employ the method-function approach?

pyrigi/_utils/_doc.py
```diff
@@ -25,19 +30,221 @@ def copy_doc(
...
+    if m:
+        prefix, assignment, func, var, rest_args, suffix = m.groups()
+        if func in class_methods:
+            if rest_args:
+                return f"{prefix}{assignment}{var}.{func}({rest_args}){suffix}"
+            else:
+                return f"{prefix}{assignment}{var}.{func}(){suffix}"
+
+    # Fallback: re.sub for wrapped calls
+    if not re.search(r"(?:>>>|\.\.\.)\s", line):
```
MatteoGallet: Maybe change this into

```python
if not re.search(r"^\s*(?:>>>|\.\.\.)\s", line):
```

so that the ``...`` is only searched after ``>>>`` and not at any point in the line?

test/utils/test__doc.py
MatteoGallet, line 1: Maybe add the following tests?

```python
def test_transform_doctest_func_to_method_wrapped_call():
    # Exercises the re.sub fallback, not the re.match primary path.
    line = ">>> len(list(all_k_extensions(G, 0)))"
    result = _transform_doctest_func_to_method(line, {"all_k_extensions"})
    assert result == ">>> len(list(G.all_k_extensions(0)))"
```

```python
def test_func_to_method_doc_none_docstring():
    assert func_to_method_doc(None, {"is_rigid"}) is None
```

```python
def test_transform_doctest_func_to_method_continuation_line_unchanged():
    line = "...            [1, 4], [2, 3], [2, 4], [3, 4]])"
    result = _transform_doctest_func_to_method(line, {"is_rigid"})
    assert result == line
```

doc/development/cheatsheet.md
Legersky: The links did not work. I agree it would be better on a separate PR, but for simplicity I think it is also fine to leave it here.

pyrigi/_utils/_doc.py
Legersky: I guess this depends whether the exactly same code will be used also for `Framework` or something else.

test/utils/test__doc.py
```diff
@@ -0,0 +1,145 @@
...
+    >>> is_linked(H, 1,3)
+    True
+    """
+
+    transformed = func_to_method_doc(doc, GRAPH_METHODS)
+
+    assert ">>> H.is_linked(1,7)" in transformed
+    assert ">>> H.is_linked(0,3)" in transformed
+    assert ">>> H.is_linked(1,3)" in transformed
+    assert ">>> is_linked(" not in transformed
```
Legersky: Is there a reason not to compare to result to the whole string transformed as we expect? That would test also that nothing else went wrong, besides the correct transformations of functions to methods.

Legersky:
After merging this, we might consider adding some basic automatic check that the examples in functions adhere to the new standard.

doc/development/internal_reference/graph/index.md
```diff
@@ -0,0 +1,44 @@
...
+   >>> len(list(all_k_extensions(g, 1, 2, only_non_isomorphic=True)))
+   ```
+   Chained calls like `graphs.Diamond().all_k_extensions(...)` are not caught
+   by the regex and will remain in method-style inside the function docstring.
+
+2. **Outer wrappers are allowed** (`list`, `len`, `type`, `sorted`, etc.), as
+   long as no argument to the class method is itself a function call:
+   ```python
+   >>> type(all_extensions(G))          # OK — no nested call in args
+   >>> len(list(all_k_extensions(G, 0)))  # OK — linear args
```
Legersky: This should replace the corresponding part here: https://pyrigi.github.io/PyRigi/development/howto/package_structure.html#graph-functionality

Legersky:
In the compiled documentation, the first argument is apparently added automatically at the end of Parameters: 
image
Is there any easy to move it to the beginning instead?

Legersky:
Any idea why it displays type `Vertex` in method style (which is good):
image
But `Hashable` in function style (not so good):
image

?

pyrigi/_utils/_doc.py
Legersky: Maybe I am missing something, but some lines in our examples do start with `...`.

Legersky:
Examples in 
`pyrigi.graph._sparsity.sparsity`
`pyrigi.graph._export.export.to_tikz`
 are not transformed to function style.

The links in `pyrigi.graph._other.apex` go to methods instead of functions.

The links in `pyrigi.graph._other.separating_set.is_stable_separating_set` do not work and the example in `pyrigi.graph._constructions.constructions.cone` needs to be decoupled.

Legersky:
The internal documentation for NAC-colorings is not displayed.

pyrigi/_utils/_doc.py
MatteoGallet: I agree. But maybe we do want to have it available for `Framework` as well, right?

test/utils/test__doc.py
khantnyarhein: I have updated the test file with mentioned tests added

MatteoGallet: Thanks!

test/utils/test__doc.py
khantnyarhein: I have updated the test file with updated check against the whole transformed string in the commit - Resolved some PR comments

doc/development/internal_reference/graph/index.md
khantnyarhein: I've replaced the part in graph-functionality, however, I haven't removed it in [doc/development/internal_reference/graph/index.md](https://github.com/PyRigi/PyRigi/pull/394/files/b35c049fb24453f11d5204ccd2e9828c6b3e59cd#diff-3bf6c46975dc91c28f2aecc0fda6759f0a874ff7a4d35ece0527b871b78c5356)

khantnyarhein:
> Any idea why it displays type `Vertex` in method style (which is good): image But `Hashable` in function style (not so good): image
> 
> ?

I've fixed it by adding `from __future__ import annotations` in the files which are in scope of this migration.

khantnyarhein:
> Examples in
> `pyrigi.graph._sparsity.sparsity`
> `pyrigi.graph._export.export.to_tikz`
> are not transformed to function style.

The reason they were not migrated seems to be because of the raw strings (r"""..."""). I've added the case and migrated the files again. However, the migration seemed to have missed a few `to_tikz()` instances but @MatteoGallet had them updated in the last commit.

khantnyarhein:
> The links in `pyrigi.graph._other.apex` go to methods instead of functions.

The migration script doesn't convert `:meth: references` with a class prefix (e.g. `:meth:~.Graph.is_k_vertex_apex`) since those can't be safely auto-converted - the same pattern can also be used for intentional method links. The four alias docstrings in `apex.py` were updated manually to `:func:`~.is_k_vertex_apex etc. This should be safe because all referenced functions are defined in the same file.

khantnyarhein:
> The links in `pyrigi.graph._other.separating_set.is_stable_separating_set` do not work and the example in `pyrigi.graph._constructions.constructions.cone` needs to be decoupled.

`:meth:`Graph.is_stable_set used `~Graph.` Fixed to `:func:`~.is_stable_set and `:func:`.is_separating_set. It should be the same case as my previous comment.
`Graph([(0,1)]).cone()` was decoupled to `g = Graph([(0,1)]); G = cone(g)`.

MatteoGallet:
> In the compiled documentation, the first argument is apparently added automatically at the end of Parameters: image Is there any easy to move it to the beginning instead?

I tried to address this issue in commit 0bb6e7302d925ddbe870f244d8d7eaf3d7f01923.

pyrigi/_utils/_doc.py
MatteoGallet: Sorry, the explanation was incorrect: this change restricts catching `...` only at the beginning of the line. This is now done in commit 8c9039da88d6dcf7fa136d5c15658bb61bbb0648.

pyrigi/_utils/_doc.py
Legersky: Yes, in another PR.

doc/development/howto/package_structure.md
```diff
@@ -69,12 +69,44 @@ which is of type {class}`networkx.Graph` and is always called `graph`.
...
+invert them to method-style at import time:
+
+1. **No chained instantiation.** Assign the graph to a variable first:
+   ```python
+   >>> g = graphs.Diamond()
+   >>> len(list(all_k_extensions(g, 1, 2, only_non_isomorphic=True)))
+   ```
+   Chained calls like `graphs.Diamond().all_k_extensions(...)` are not caught
+   by the regex and will remain in method-style inside the function docstring.
+
```
Legersky: It is the other way around, isn't it?

doc/development/howto/package_structure.md
```diff
@@ -69,12 +69,44 @@ which is of type {class}`networkx.Graph` and is always called `graph`.
...
+2. **Outer wrappers are allowed** (`list`, `len`, `type`, `sorted`, `print`, etc.), as
+   long as no argument to the class method is itself a function call:
```
Legersky: Should "class method" be "function"?

doc/development/internal_reference/graph/index.md
Legersky: Please, remove it from `graph/index.md` to avoid duplication.

doc/development/internal_reference/graph/nac.md
```diff
@@ -0,0 +1,9 @@
+# NAC Colorings
+
+```{eval-rst}
+.. automodule:: pyrigi.graph._flexibility.nac
+   :members:
+   :imported-members:
+   :undoc-members:
+   :show-inheritance:
+```
```
Legersky, line 98: The module `nac` has submodules that should be displayed in the internal reference. I think for now all of them can be on the same page.

Legersky:
When this is finished and merged, I will create a PR extending this functionality to frameworks (I have a draft ready on a local branch).

pyrigi/graph/_sparsity/_pebble_digraph.py
MatteoGallet, line 1: The methods `K`, `L`, `K_and_L`, `redirect_edge_to_head`, and `dfs` got `graph` among their parameters, which should not be there.

pyrigi/graph/_rigidity/realization_counting.py
MatteoGallet, line 1: The methods `_number_of_sphere_realizations_min_rigid_dim_2_rec` and `_number_of_plane_realizations_min_rigid_dim_2_rec` have `graph` among their parameters, which should not be there.

pyrigi/graph/_flexibility/nac/check.py
MatteoGallet, line 107: Here the parameter `graph` is not mentioned.

pyrigi/graph/_rigidity/realization_counting.py
Legersky: ✅

pyrigi/graph/_flexibility/nac/check.py
Legersky: ✅

pyrigi/graph/_sparsity/_pebble_digraph.py
Legersky: ✅

pyrigi/_utils/_doc.py
```diff
@@ -25,19 +30,273 @@ def copy_doc(
...
+        if new_line != line:
+            result.append(new_line)
+            continue
+
+        # 3. :func: -> :meth: for bare names
+        line = _transform_func_to_meth_refs(line, class_methods)
+
+        # 4. "this function" / "the function" / "The function"
+        line = re.sub(r"\bthis function\b", "this method", line)
+        line = re.sub(r"\bthe function\b", "the method", line)
```
MatteoGallet, line 293: Maybe add:

```python
        line = re.sub(r"\bThis function\b", "This method", line)
```

pyrigi/_utils/_doc.py
```diff
@@ -25,19 +30,273 @@ def copy_doc(
...
+        stripped = line.strip()
+
+        # Track when we enter the Parameters section and its dashes line.
+        if stripped == "Parameters":
+            in_params = True
+            after_dashes = False
+        elif in_params and not after_dashes and re.match(r"^-+$", stripped):
+            after_dashes = True
+        elif in_params and after_dashes:
+            # Skip a bare 'graph:' or 'framework:' entry (no description follows).
```
MatteoGallet, line 266: Maybe we should document that functions that get wrapped may have  a bare `graph` or `framework` parameter, which gets removed, but those parameters should not be followed by descriptions, since they would remain?

doc/development/howto/package_structure.md
```diff
@@ -69,12 +69,45 @@ which is of type {class}`networkx.Graph` and is always called `graph`.
...
-and not functions taking {class}`networkx.Graph`.
+The decorator `@copy_doc` not only copies the docstring but also converts
+function-style examples to method-style at import time.
+Therefore, docstrings should be written in **function-style**, referring to the
+graph as "the graph" rather than by the parameter name `graph`.
+
+(writing_examples)=
+### Writing examples for `copy_doc`-wrapped functions
+
+Examples in function-style docstrings must follow two rules so `copy_doc` can
```
MatteoGallet, line 84: "three rules" ?

pyrigi/_utils/_doc.py
Legersky: :heavy_check_mark:

pyrigi/_utils/_doc.py
Legersky: :heavy_check_mark:

MatteoGallet (approved):

PR #396, opened 2026-06-04
Code: Speedup NAC coloring search by using adjacency matrix vs NetworkX graph

MatteoGallet:


Legersky:
@khantnyarhein If you have time, could you provide some timing comparison before and after the change using your benchmarking tool as we discussed? You can use for instance the class of graphs with `2n-3` edges.
However, this is not necessary before the PR being merged as @MatteoGallet did some brief tests.

Lastaapps:
@MatteoGallet is it the same snippet we discusseded?

MatteoGallet:
> @MatteoGallet is it the same snippet we discusseded?

Yes!

Lastaapps (approved):

PR #405, opened 2026-06-26
Test: Added test verifying parameter swap detection

Legersky:


khantnyarhein (approved):
I think it looks good.

